PR #35, opened 2022-11-18
feat: Redesign DSL and strategy implementation.

jimsynz:


lib/ash_authentication.ex
```diff
@@ -1,101 +1,23 @@
...
-          store revocation information. This option is the name of an Ash
-          Resource which has the `AshAuthentication.TokenRevocation` extension
-          present.
-          """
-        ]
-      ]
-    }
-  ]
+  require AshAuthentication.Dsl
+  import AshAuthentication.Dsl
```
zachdaniel, line 3: importing should require, IIRC

lib/ash_authentication.ex
```diff
@@ -1,101 +1,23 @@
...
-  This package assumes that you have [Phoenix](https://phoenixframework.org/) and
-  [Ash](https://ash-hq.org/) installed and configured.  See their individual
-  documentation for details.
+  This package assumes that you have [Ash](https://ash-hq.org/) installed and
+  configured.  See the Ash documentation for details.
 
-  Once installed you can easily add support for authentication by configuring one
-  or more extensions onto your Ash resource:
+  Once installed you can easily add support for authentication by configuring
+  one or more extensions onto your Ash resource:
```
zachdaniel, line 15: should this now just say something like "by configuring the AshAuthentication extension on your resource"

lib/ash_authentication.ex
```diff
@@ -171,42 +103,40 @@ defmodule AshAuthentication do
...
+      iex> build_user(id: "ce7969f9-afa5-474c-bc52-ac23a103cef6") |> user_to_subject()
+      "user?id=ce7969f9-afa5-474c-bc52-ac23a103cef6"
 
-  @doc """
-  Return a subject string for an AshAuthentication resource.
   """
-  @spec resource_to_subject(Resource.record()) :: subject
-  def resource_to_subject(record) do
+  @spec user_to_subject(Resource.record()) :: subject
+  def user_to_subject(record) do
```
zachdaniel, line 136: Should we be using the term `user` here? I think we may want to use the term `actor`, since theoretically something like a service account could could be authenticating.

jimsynz: I went back and forwards on this a few times, but ultimately I went back to `user` as being the common case and less confusing than other options - most of which already have meanings in Ashland.

zachdaniel: Talked off line, agreed to stick with user as a specific concept to `ash_authentication`. i.e "its not an actor until you use it as an actor in your ash resources".

zachdaniel (commented):
LFG 😎

zachdaniel (approved):

PR #46, opened 2022-11-24
improvement(Confirmation): Confirmation is not a strategy.

jimsynz:
Well, actually it is - but conceptually it's not.


lib/ash_authentication/add_ons/confirmation.ex
```diff
@@ -0,0 +1,140 @@
+defmodule AshAuthentication.AddOn.Confirmation do
+  import AshAuthentication.Dsl
+
+  @moduledoc """
+  Confirmation support.
+
+  Sometimes when creating a new user, or changing a sensitive attribute (such as
+  their email address) you may want to for the user to confirm by way of sending
```
rellen, line 8: I think there's a typo here???
```suggestion
  their email address) you may want wait to for the user to confirm by way of sending
```

lib/ash_authentication/add_ons/confirmation/transformer.ex
```diff
@@ -49,8 +49,8 @@ defmodule AshAuthentication.Strategy.Confirmation.Transformer do
       [_ | _] ->
         {:error,
          DslError.exception(
-           path: [:authentication, :strategies, :confirmation],
-           message: "Multiple confirmation strategies are not supported"
+           path: [:authentication, :add_ons, :confirmation],
+           message: "Multiple confirmation add ons are not supported"
```
zachdaniel, line 53: unrelated to this PR, but why not? Is it routing? If we put the add_on name in the route, we could support any number of confirmations, right?

jimsynz: yeah, that occurred to me too. seems like an arbitrary decision I made at the start

lib/ash_authentication/add_ons/confirmation.ex
jimsynz: Nice catch!

lib/ash_authentication/add_ons/confirmation.ex
```diff
@@ -0,0 +1,140 @@
...
+          name: :confirm
+        }
+
+  @doc """
+  Generate a confirmation token for a changeset.
+
+  This will generate a token with the `"act"` claim set to the confirmation
+  action for the strategy, and the `"chg"` claim will contain any changes.
+
+  FIXME: The "chg" claim should encrypt the contents of the changes so as to not
```
zachdaniel, line 121: we should make a GH issue for this, so this isn't lost.

zachdaniel (approved):

rellen (approved):
just one doco typo found.  I am not across the particulars to review the code changes

lib/ash_authentication/add_ons/confirmation.ex
jimsynz: :+1:

PR #70, opened 2022-12-08
improvement: better transaction workaround

zachdaniel:
The `DataLayer.transaction` callback is a bit low level, better to use this workaround for bulk deleting in a transaction. Before too long we'll have bulk destroys and we'll be able to use that instead :)


zachdaniel:
I've changed my mind on this implementation :)

PR #75, opened 2022-12-08
improvement: use `Code.ensure_compiled!` which "demands" that the com…

zachdaniel:
…piler wait for X module


jimsynz (approved):
That'll do

PR #80, opened 2022-12-08
Improvement(Jwt)!: use secrets for jwt config

jimsynz:
Use the `AshAuthentication.Secret` behaviour, rather than asking the user to explicitly set it in their application environment.


zachdaniel (approved):
had a chat about application env vs system env, otherwise LGTM

PR #85, opened 2022-12-10
Quick fixes

zimt28:
I'm currently trying to switch to `ash_authentication` and will push fixes to small issues I see along the way. Not sure if this should be a PR as it's not nice in any way but at least I won't forget about anything this way.

### DIFF
--- a/lib/ash_authentication/strategies/password/sign_in_preparation.ex
+++ b/lib/ash_authentication/strategies/password/sign_in_preparation.ex
@@ -30,13 +30,11 @@ defmodule AshAuthentication.Strategy.Password.SignInPreparation do
     |> Query.after_action(fn
       query, [record] ->
         password = Query.get_argument(query, strategy.password_field)
+        hashed_password = Map.get(record, strategy.hashed_password_field)
 
-        if strategy.hash_provider.valid?(
-             password,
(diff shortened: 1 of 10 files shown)


lib/ash_authentication/strategies/password/sign_in_preparation.ex
```diff
@@ -30,8 +30,9 @@ defmodule AshAuthentication.Strategy.Password.SignInPreparation do
     |> Query.after_action(fn
       query, [record] ->
         password = Query.get_argument(query, strategy.password_field)
+        hashed_password = Map.get(record, strategy.hashed_password_field)
```
jimsynz, line 33: Good spotting!

lib/ash_authentication/token_resource/transformer.ex
```diff
@@ -60,7 +60,7 @@ defmodule AshAuthentication.TokenResource.Transformer do
              writable?: true
            ),
          {:ok, dsl_state} <-
-           maybe_build_attribute(dsl_state, :created_at, :utc_datetime_usec,
+           maybe_build_attribute(dsl_state, :inserted_at, :utc_datetime_usec,
```
jimsynz, line 63: Don't sleep and code kids.

PR #95, opened 2022-12-15
improvement: docs updates

zachdaniel:
The numbers in the names mess with AshHq. We could manually do the guides list in the DocIndex, but this was faster. Happy to walk it back. We also now support regexes in the code modules, so I updated it to include additional docs.


zachdaniel:
Once we get a release w/ the guide names improved, I'll add the libraries to the ash_hq db and it should index 👍🏻

zachdaniel:
If you're good with it, I'll merge it (or you can, all the same to me)

PR #102, opened 2022-12-19
improvement: Allow the strategy name to be passed for password validations and changes.

jimsynz:
After #89 was merged folks were no longer able to use `AshAuthentication.Strategy.Password.HashPasswordChange` and `AshAuthentication.Strategy.Password.PasswordConfirmationValidation` in their own actions.  This change fixes this issue by allowing the name of the strategy to be passed in in the changeset context.


lib/ash_authentication/strategies/password/hash_password_change.ex
```diff
@@ -4,6 +4,18 @@ defmodule AshAuthentication.Strategy.Password.HashPasswordChange do
...
   Uses the configured `AshAuthentication.HashProvider` to generate a hash of the
   user's password input and store it in the changeset.
+
+  You can use this change in your actions where you want to change the user's
+  password.  If you're not using one of the actions generated by the password
+  strategy then you'll need to manually pass the strategy name in the changeset
+  context.  Eg:
+
+  ```elixir
+  Changeset.new(user, %{})
```
zachdaniel, line 14: I think we should show an example here of passing the context statically in a named action, as that is more idiomatic in this case.

lib/ash_authentication/strategies/password/password_confirmation_validation.ex
```diff
@@ -2,7 +2,20 @@ defmodule AshAuthentication.Strategy.Password.PasswordConfirmationValidation do
...
+  This check is only performed when the `confirmation_required?` DSL option is
+  set to `true`.
+
+  You can use this validation in your own actions where you want to validate
+  that the password and the password confirmation arguments match.  If you're
+  not using one of the actions generated by the password strategy then you'll
+  need to manually pass the strategy name in the changeset context.  Eg:
+
+  ```elixir
+  Changeset.new(user, %{})
```
zachdaniel, line 14: Same feedback as above.

zachdaniel (approved):
Small feedback

lib/ash_authentication/strategies/password/hash_password_change.ex
jimsynz: You mean as in `api.action(changeset, context: %{strategy_name: :password})`?  If so we can do that for the change, but not the validation as it's not passed the action context.

zachdaniel: I mean like showing them an action definition that sets the context.

```elixir
update :change_password do
  change set_context(%{strategy_name: :password})
end
```

jimsynz: oooooh right. yeah, of course.

minibikini: @zachdaniel I tried the snippet above and got this error:

```
== Compilation error in file lib/table_hero/accounts/resources/user.ex ==
** (UndefinedFunctionError) function Ash.Changeset.set_context/1 is undefined or private. Did you mean:

      * set_context/2

    (ash 2.4.27) Ash.Changeset.set_context(%{strategy_name: :password})
    lib/table_hero/accounts/resources/user.ex:40: (module)
```

Which makes sense because `set_context/2` requires both a context and a map. 

Question: how do I get a context inside the update macro, so I'd be able to provide it to `set_context/2`?

zachdaniel: The example I was providing was inside of an action declaration, what exactly did you try?

zachdaniel: @minibikini in that context its using a declarative "change". There is a builtin change called `set_context/1` that you can use in an action, so with this in your resource:

```elixir
update :change_password do
  ...
  change set_context(%{strategy_name: :password})
  change HashPasswordChange
end
```

You can call `YourResource |> Ash.Changeset.for_update(:change_password, ....) |> Api.update()` and it will call those changes in order (sort of like plugs).

minibikini: I was trying something like this:

```elixir
defmodule MyApp.Accounts.User do
  actions do  
    update :password do
      ...
      change Changeset.set_context(%{strategy_name: :password})
      change AshAuthentication.Strategy.Password.HashPasswordChange      
    end   
  end
end

Comment (minibikini) [reply lib/ash_authentication/strategies/password/hash_password_change.ex:14]: I've tried this
```elixir
change set_context(%{strategy_name: :password})
```

and got this error:

```
(CompileError) function set_context/1 imported from both Ash.Resource.Preparation.Builtins and Ash.Resource.Change.Builtins, call is ambiguous

Stacktrace:
  │ (ash 2.4.26) expanding macro: Ash.Resource.Dsl.Actions.Update.Actions.Changes.Change.change/1
  │ lib/my_app/accounts/resources/user.ex:40: MyApp.Accounts.User (module)
  │ (ash 2.4.26) expanding macro: Ash.Resource.Dsl.Actions.Update.update/2
  │ lib/my_app/accounts/resources/user.ex:20: MyApp.Accounts.User (module)
  │ (ash 2.4.26) expanding macro: Ash.Resource.Dsl.actions/1
  │ lib/my_app/accounts/resources/user.ex:9: MyApp.Accounts.User (module)
```

zachdaniel: 🤔 I've never seen that issue before. Can I see the whole resource?

zachdaniel: wow, sure enough.

zachdaniel: I'm seeing the same issue.

zachdaniel: For now, just do this:

```elixir
change fn changeset, _ -> 
  Ash.Changeset.set_context(changeset, %{strategy_name: :password})
end
```

I'll have to fix the way we do imports to prevent the clash between preparations and changes. Sorry about that.

minibikini: Here is the whole resource https://gist.github.com/minibikini/8d2ff6a69de10cc94f4cb697eaca227c

zachdaniel: Okay, latest version of Ash has this fixed

minibikini: Just tested the branch in my project - it works!

PR #115, opened 2023-01-09
improvement: Set Ash actor and tenant when executing internal plugs.

### LINKED ISSUE #114: Strategy plugs need to honour the actor and tenant when calling actions.

jimsynz:
Closes #114.

### DIFF
--- a/lib/ash_authentication/plug/router.ex
+++ b/lib/ash_authentication/plug/router.ex
@@ -6,7 +6,7 @@ defmodule AshAuthentication.Plug.Router do
   Used internally by `AshAuthentication.Plug`.
   """
 
-  alias AshAuthentication.{Info, Strategy}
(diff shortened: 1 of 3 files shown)


zachdaniel:
I don't think its a good idea to use the context based approach to setting actor/tenant really. I advise against it in the docs generally, and things can get weird (i.e if somehow code happens in a different process, the actor isn't automatically carried over)

jimsynz:
> I don't think its a good idea to use the context based approach to setting actor/tenant really. I advise against it in the docs generally, and things can get weird (i.e if somehow code happens in a different process, the actor isn't automatically carried over)

Okie dokie. I'll do the slightly more longwinded method.

jimsynz:
@zachdaniel updated.

zachdaniel (approved):

PR #116, opened 2023-01-11
feat: Add token-required-for-authentication feature.

jimsynz:
* Adds the `require_token_presence_for_authentication?` DSL option to the Authentication extension which when enabled changes the following behaviour:
  1. The `store_in_session` plug will store the user's token rather than their subject in the session.
  2. The `retrieve_from_session` plug will look for a stored token in the session rather than a subject and ensure that it's present in the `TokenResource`.
  3. The `retrieve_from_bearer` plug will ensure that the token is present in the `TokenResource`.
* Adds the `get_token` action to the `TokenResource`.


lib/ash_authentication/dsl.ex
```diff
@@ -69,7 +69,7 @@ defmodule AshAuthentication.Dsl do
       %Section{
         name: :authentication,
         describe: "Configure authentication for this resource",
-        modules: [:api],
+        # modules: [:api],
```
zachdaniel, line 72: ?

PR #119, opened 2023-01-11
improvement: add policy utilities and accompanying guide

zachdaniel:
fix: improve some error message/validation logic

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -5,7 +5,5 @@ config :mime, :types, %{
 }
 
 config :ash, :utc_datetime_type, :datetime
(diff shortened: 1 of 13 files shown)


zachdaniel:
Not sure whats up with that build failure.

jimsynz:
it's this: https://github.com/elixir-mint/castore/issues/50

zachdaniel:
I think we may need to include the compile time config files in our dep recompilation cache, haven't looked into it too much yet. But `compile_env` is used by Ash in some cases, and that is causing this error.

PR #133, opened 2023-01-13
improvement(TokenResource)!: Store the token subject in the token resource.

jimsynz:
This is a breaking change because you may have to delete tokens in your database so that you can avoid the non-null constraint on subject.


zachdaniel:
Since there are people using the package now, I think we should create a `documentation/topics/upgrading.md` or similar that contains a heading per version with breaking change, explaining what needs to change.

jimsynz:
@zachdaniel I have added an upgrading guide.

zachdaniel (approved):

PR #134, opened 2023-01-13
docs: Add a small warning about defining the failure template.

### LINKED ISSUE #118: Failure template error is confusing
When following the default setup, if something goes wrong with authentication in some way, we currently "blow up" (if following the getting started guide). I think what we should do is, instead of rendering `failure.html` we should redirect to the sign in url with a flash message in our example auth controller.

jimsynz:
Closes #118.

### DIFF
--- a/documentation/tutorials/integrating-ash-authentication-and-phoenix.md
+++ b/documentation/tutorials/integrating-ash-authentication-and-phoenix.md
@@ -107,7 +107,7 @@ and another for your `:browser` pipeline). Let's define a version for a browser
 client:
 
 > Remember to define an appropriate template in `failure.html.heex` for your
-> controller.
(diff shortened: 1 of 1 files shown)


documentation/tutorials/integrating-ash-authentication-and-phoenix.md
```diff
@@ -106,6 +106,9 @@ You can define multiple versions if required (eg one for your `:api` pipeline
 and another for your `:browser` pipeline). Let's define a version for a browser
 client:
 
+> Remember to define an appropriate template in `failure.html.heex` for your
+> controller.
```
zachdaniel, line 110:
```suggestion
> controller. Alternatively, you could redirect with a flash message.
```

PR #138, opened 2023-01-16
Fix `signing_secret` examples

zimt28:


jimsynz (commented):
Am I missing something? [`Application.fetch_env/2`](https://hexdocs.pm/elixir/Application.html#fetch_env/2) does return an ok tuple.

zimt28:
@jimsynz You aren't. I used `Application.fetch_env!/2` and had to return the tuple manually, but `fetch_env/2` already does that of course.. Shouldn't have worked until 3am :)

PR #151, opened 2023-01-18
fix(Password): validate fields using both methods of allowing nil input.

### LINKED ISSUE #150: Shouldn't require `allow_nil_input [...]` for a field that itself is `allow_nil? true`
I was attempting to overwrite the `:register_with_password` action to expand the arguments and fields, but get the following error even though the `:hashed_password` attribute includes `allow_nil?: true`:
```shell
Expected the action `:register_with_password` to allow nil input for the field `:hashed_password`
```

For more info, see original Discord discussion:
https://discord.com/channels/711271361523351632/1050299714408550400/1065296028405878805

jimsynz:
Closes #150.


zachdaniel (approved):
🥳

PR #154, opened 2023-01-20
feat: Add support and documentation for custom strategies.

jimsynz:


zachdaniel:
Love this. Have done a light review of the changes, will want to look into it a bit more deeply later. Very pleased with defining all the built in strategies using the same tools for custom strategies. That pattern worked well for ash/spark because it ensures that anything we build into core also adds capabilities for custom strategies. Things that stood out to me and I’ll want to look at more: the EEx.eval and the fact that transformer order is currently not configurable for strategies (may not be an issue at all).

jimsynz:
The `EEx` stuff is only in the dev server, so not part of the package per se. I thought about transformer ordering, but it all got too difficult and I didn't need it yet. I was thinking it would be cool if we could change the way entity transformers work so that they could transform more than just the entity but the whole DSL state.

zachdaniel:
It's only on the dev server but I assume it's what people would need to do to give their custom strategies UI components?

As for entity transformers transforming the whole DSL, it adds some complexity to the problem of when things are transformed which is already pretty complex. I could support the addition of an entity adding a new transformer that runs with the other transformers though? We'd need "opts" for transformers though...lol

jimsynz:
> It's only on the dev server but I assume it's what people would need to do to give their custom strategies UI components?

Nope, it's just so that I can do manual testing in dev.

> As for entity transformers transforming the whole DSL, it adds some complexity to the problem of when things are transformed which is already pretty complex. I could support the addition of an entity adding a new transformer that runs with the other transformers though? We'd need "opts" for transformers though...lol

Yeah, it just felt weird to reinvent the transformer mechanism for this change.

zachdaniel:
Agreed,the ability to have transformers get added automatically based on entities being added solve for it. Alternatively, instead of a `transform` callback, you could have a `transformers` callback that returns the transformers. Then you don't have to recreate anything

zachdaniel:
What would the render look like in the real world?

jimsynz:
> What would the render look like in the real world?

Well they'd have to make their own form or liveview or API endpoint.

zachdaniel:
I thought we did something where we loop over each strategy to see if it has a form and render that. Or is that only for the specific built in strategies? I see what you mean now though, the render function isn't part of the behaviour just something you're using for tests

jimsynz:
> I thought we did something where we loop over each strategy to see if it has a form and render that. Or is that only for the specific built in strategies? I see what you mean now though, the render function isn't part of the behaviour just something you're using for tests

We do that in `ash_authentication_phoenix` but `ash_authentication` itself doesn't provide any UI at all.

zachdaniel:
Right. And custom strategies don't have a way to be included in that process currently?

jimsynz:
Nope, although I am working on a PR to `ash_authentication_phoenix` which allows you to add custom strategy components.

zachdaniel:
Nice :)

PR #164, opened 2023-01-28
improvement: add metadata declarations to actions that have a `token`

zachdaniel:
Nothing in ash core uses these static metadata declarations, but they are used by API extensions to know that they should potentially expose action-specific metadata. Currently only AshGraphql will use this information


zachdaniel:
The reason for this is because if a user wants to sign in over a graphql and then retrieve the token on the client, ash_graphql needs a way to know to add that to the type.

jimsynz:
It needs to go anywhere that `GenerateTokenChange` is used - that means in OAuth also.

zachdaniel:
But those actions you define yourself, right? So you'd add the metadata yourself if you wanted to. Its unlikely that people would hook the oauth actions up to an api extension, but they potentially could.

jimsynz:
> But those actions you define yourself, right? So you'd add the metadata yourself if you wanted to. Its unlikely that people would hook the oauth actions up to an api extension, but they potentially could.

DOH!

jimsynz (approved):
big brain

PR #183, opened 2023-02-08
improvement(Strategy.Custom): handle custom strategies as extensions.

jimsynz:
This means that users can add their own extensions to their resources which patch the strategy (and add ons) DSLs.


jimsynz:
This PR is just waiting on a Spark release, then it will be ready to go.

zachdaniel (approved):

PR #198, opened 2023-02-17
Fix sign in preparation without identity resource

hwuethrich:
I noticed sign-in does throw an error when registration and identity resource are disabled. The test might need some work, but I couldn't override `registration_identity` in the test like this:

```elixir
strategy = %{strategy | registration_enabled?: false, identity_resource: false}
```


jimsynz (approved):
nice find

jimsynz:
@hwuethrich thanks so much for your contribution.

PR #213, opened 2023-02-25
fix: respect `identity_relationship_user_id_attribute` on `Strategy.OAuth2.IdentityChange`

quartz55:
Noticed this when I changed from the default `user_id` to e.g. `account_id`:
```ex
[error] Something went wrong in authentication

activity: {:oauth2, :callback}

reason: ** (Ash.Error.Invalid) Input Invalid

* argument account_id is required
  (ash 2.5.16) lib/ash/changeset/changeset.ex:860: anonymous fn/2 in Ash.Changeset.require_arguments/2
  (elixir 1.14.3) lib/enum.ex:2468: Enum."-reduce/3-lists^foldl/2-0-"/3
  (ash 2.5.16) lib/ash/changeset/changeset.ex:706: Ash.Changeset.do_for_action/4
  (ash_authentication 3.7.9) lib/ash_authentication/user_identity/actions.ex:29: AshAuthentication.UserIdentity.Actions.upsert/2
  (ash_authentication 3.7.9) lib/ash_authentication/strategies/oauth2/identity_change.ex:33: anonymous fn/3 in AshAuthentication.Strategy.OAuth2.IdentityChange.do_change/2
...
  ```

### DIFF
--- a/lib/ash_authentication/strategies/oauth2/identity_change.ex
+++ b/lib/ash_authentication/strategies/oauth2/identity_change.ex
@@ -29,20 +29,22 @@ defmodule AshAuthentication.Strategy.OAuth2.IdentityChange do
   defp do_change(changeset, strategy) do
     changeset
     |> Changeset.after_action(fn changeset, user ->
-      strategy.identity_resource
-      |> UserIdentity.Actions.upsert(%{
-        user_info: Changeset.get_argument(changeset, :user_info),
-        oauth_tokens: Changeset.get_argument(changeset, :oauth_tokens),
-        strategy: Strategy.name(strategy),
-        user_id: user.id
-      })
-      |> case do
-        {:ok, _identity} ->
-          user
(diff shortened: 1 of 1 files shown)


jimsynz:
Nice catch on this bug.  Can you change it so that it uses the generated functions in `AshAuthentication.UserIdentity.Info` rather than reaching into the strategy directly?  Thanks!

quartz55:
Sure thing! Had to change the structure around though

jimsynz:
Thanks @quartz55. For some reason the CI is not running on this PR which is just weird.  Regardless, `Info.user_identity_user_id_attribute_name/1` can return `:error` so I'm getting a dialyzer failure because there needs to be an `else` clause for it.

quartz55:
Oh right, my bad, forgot to run `mix ci` that time. Everything's green now 👍

PR #214, opened 2023-02-26
docs: Updates confirmation docs with sender and identities

britton-jb:
Small hiccup in the docs for adding confirmation - not a huge blocker, but could slow folks down on adoption.


jimsynz:
Thanks for the contribution!

PR #219, opened 2023-03-02
feat!: Configure accepted fields on register 

zachdaniel:
also fixes some .formatter.exs issues and standardizes it using spark tools

This is a breaking change because users of previous versions may be relying on this
behavior. To upgrade, add `register_action_accept [:additional, :fields]` to your
password strategy. This prevents users accidentally adding common security flaw which
is accepting more than intended on a given action.


zachdaniel:
still working on this

jimsynz (approved):
ship it

PR #254, opened 2023-04-03
improvement: update spark

zachdaniel:


jimsynz:
You know that @dependabot will handle this eh?

PR #359, opened 2023-07-14
fix: ensure that the `current_` atom exists at compile time.

### LINKED ISSUE #358: Non existent atom error
```elixir
** (exit) an exception was raised:
    ** (ArgumentError) errors were found at the given arguments:

  * 1st argument: not an already existing atom

        :erlang.binary_to_existing_atom("current_user", :utf8)
```

Nothing in my application yet refers to `:current_user` and so therefore we get an error about a non existing atom. I think perhaps we should make these atoms at compile time for each resource.

jimsynz:
Fixes #358.


rellen:
👍

PR #389, opened 2023-08-16
fix: correct spec for `Jwt.token_for_user`

vonagam:
Should be `map` instead of `%{}` for `extra_claims`. Was getting "has no local return" from dialyzer because of this. (This is the only place where `%{}` is used instead of `map` in spec.)


jimsynz:
Thanks for the contribution! 🚀

PR #424, opened 2023-09-18
fix: only use sign in token expiration for sign in tokens

zachdaniel:


mindok:
@jimsynz This broke token based auth for graphql / api. Do I need to change a setting somewhere?

PR #430, opened 2023-09-21
improvement(TokenResource)!: rename generated `created_at` attribute to `inserted_at`.

jimsynz:
If you are using the token resource this is a **breaking change**.  You will need to generate new migrations using:

    mix ash_postgres.generate_migrations --name=rename_created_at_to_inserted_at

It will ask you if you are renaming `created_at` to `inserted_at` - answer yes.

This updates and closes #85.


jimsynz:
Thoughts @zachdaniel? Not sure if it's worth the major release just to change the name of an attribute.

zachdaniel:
IMO it is definitely not worth it. We can make it configurable if some people need that.

jimsynz:
:+1:

PR #442, opened 2023-09-28
Fail fast when misconfigured 🧨

dewetblomerus:
When I first followed the guide, I copy/pasted the code the read the `:redirect_uri` key, but when I typed into my config file, I typed `:redirect_url`.

The failure was completely silent with no error raised.

This change will cause the app to fail loud and clear with  the following error:

```elixir
 (KeyError) key :redirect_uri not found in: [
  client_id: "the-client-id",
  redirect_url: "http://localhost:4000/auth",
  client_secret: "the-client-secret",
  site: "the-site"
]
```


zachdaniel:

PR #471, opened 2023-10-14
Fix missing piece of module path in doc example

lawik:
I think this is a mistake. I didn't cast about for more instances of it.


zachdaniel:

PR #473, opened 2023-10-14
Change redirect_uri secret to be more flexible

lawik:
Applies to both OAuth2 and OpenID Connect.

With this the developer can provide either the full URL or the URL up to the AuthPlug path. If the suffix is already there we won't add it again.

Should make it easier to get it on the first try.


zachdaniel (approved):
Will let @jimsynz have a look also though.

lawik:
Yeah, hit it hard :)

jimsynz (approved):
amazing

PR #474, opened 2023-10-14
feat: Add Google strategy

lawik:


lawik:
I did not test this yet. I also did not figure out how to generate `documentation/dsls` which says they are generated by spark.

zachdaniel (approved):
Noice! Let's get it tested in at least one case before merging. But even with that, its almost entirely assent under the hood, so there isn't a whole lot to test from an automation standpoint.

PR #481, opened 2023-10-17
Add id as an option for sourcing uid for UserIdentity

lawik:
Not sure if this is a fix that is wanted. But I had Google OAuth fail any time I switched the identity_resource thing on in the strategy.


jimsynz (approved):
nice

PR #483, opened 2023-10-17
Change overwriting of refresh_token to not overwrite them with nil

lawik:


lawik:
Not sure if there is a preferred way to add a conditional to this type of pipeline in Ash. This is just a stab at it.

zachdaniel (approved):

PR #484, opened 2023-10-17
docs: adds guide for oauth2 strategy with google

Sam23D:
Adds a simple guide on how to configure Google sign in and registration with the oauth2 strategy


jimsynz (approved):
perfect

PR #566, opened 2024-02-22
better Plug.Router example

jlgeering:
I'm pretty sure `plug :match` and `plug :dispatch` are required when using `Plug.Router`


zachdaniel:
Yeah, I think so too, but it might also be worth just trimming the example down to only include the relevant parts for ash_authentication and exclude the rest w/ a `...` or something. I'll merge this though.

zachdaniel:

PR #570, opened 2024-02-26
docs: Mark the `hashed_password` field as private so it does not appear in any API responses

sevenseacat:
As the API extensions grab all public attributes by default - this is one that really should not be included!


jimsynz:
Thanks!!

PR #589, opened 2024-03-11
improvement: infer `api` from a resource

vonagam:
Infer `api` set by `use Ash.Resource, api: ...` in `authentication`, `token` and `user_identity` sections.

Safe to remove `required: true` from `api` options because those sections verifiers already do verify that `api` is set and valid.


jimsynz (approved):
thanks partner

PR #617, opened 2024-04-06
fix incorrect file name in tutorial

TheArrowsmith:
It's `resources/user.ex` earlier in the file so presumably it's supposed to be `resources/user.ex` here too.


zachdaniel:
The latest suggestions from Ash core are actually to not put resources in a `resources` folder. Would you mind changing the references not to include `/resources`? We're doing a docs review now to address many issues like this.

TheArrowsmith:
sure - have removed refs to `resources` and also fixed a couple of typos I spotted.

zachdaniel (approved):

zachdaniel:

PR #665, opened 2024-05-02
docs: Documentation refresh for Ash 3.0.

jimsynz:
Remaining to do next week:

- [ ] Testing topic
- [ ] Policies topic


README.md
```diff
@@ -1,50 +1,81 @@
...
 [![Hex version badge](https://img.shields.io/hexpm/v/ash_authentication.svg)](https://hex.pm/packages/ash_authentication)
+[![Hexdocs badge](https://img.shields.io/badge/docs-hexdocs-purple)](https://hexdocs.pm/ash_authentication)
+
+# Ash Authentication
+
+Welcome! Here you will find everything you need to know to get started with and use Ash Authentication. This documentation is best viewed on [hexdocs](https://hexdocs.pm/ash_authentication).
+
+AshAuthentication provides drop-in support for user authentication for users of the [Ash Framework](https://ash-hq.org). It is designed to be highly configurable, with sensible defaults covering the most common use-cases.
+
+## About the Documentation
```
zachdaniel, line 15: We've been leaving this bit out of the sub-project guides, but I'm personally on the fence about it, so its up to you :D

documentation/how-to/confirmation.md
```diff
@@ -0,0 +1,173 @@
+# Confirmation How-To
+
+This is a quick tutorial on how to configure your application to enable confirmation.
```
zachdaniel, line 3: An important bit of the docs refactor is making sure that each thing properly adheres to its "type". This does in fact appear to be a tutorial to me, and so this should probably go in `/tutorials`. 

I've found that there is actually almost zero "how-to" documentation according to diataxis in Ash. There is explanation, and tutorials, and reference, but not so much "how-to". (just a meta comment).

jimsynz:
I'm leaving the testing and policies topics as they are for now because although I have a lot I want to say, I can't get the words straight so I'm not going to hold up this PR any longer.

PR #676, opened 2024-05-12
doc: fix code example on docs

suciptoid:

PR #678, opened 2024-05-13
docs: fix `ash_domains` config example in `Get Started` section

arcanemachine:
I believe the correction in #676 was slightly off the mark. This little fix adjusts the syntax to the correct value (I think).

### DIFF
--- a/documentation/tutorials/get-started.md
+++ b/documentation/tutorials/get-started.md
@@ -109,7 +109,7 @@ Be sure to add it to the `ash_domains` config in your `config.exs`
 
 ```elixir
 # in config/config.exs
-config :my_app, ash_domains: [..., MyApp.Accounts]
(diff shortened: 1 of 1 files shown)


zachdaniel:
AFAIK the syntax you're replacing works. Perhaps they both work?

```elixir
config :ash_hq,
  ash_domains: [
    AshHq.Accounts,
    AshHq.Blog,
    AshHq.Docs,
    AshHq.Github,
    AshHq.MailingList,
    AshHq.Discord
  ]
```

This is what we do in `ash_hq` for example. Will close for now, but if I'm wrong LMK and we'll reopen :)

arcanemachine:
Sorry, that's probably just confusion on my part regarding the `config` syntax. I thought it would write the data to a different section/key.

Note to self: Test things manually first 😅

PR #681, opened 2024-05-13
docs: improve security in the 'Get Started' guide

arcanemachine:
A few small changes to the tutorial to make the sections related to secrets a little less insecure. It's still not super secure, but I think it serves as a better jumping-off point which sends users in the right direction.

- Use a bang function to ensure that the `signing_secret` will raise an exception if it is absent from the config (instead of just silently evaluating to `nil`)

- Add config example for the secret signing token

- Add warning so that users are informed about the security implications of using the basic configuration provided by the tutorial

- Rearrange the sections a bit so the `Token Generation` section comes after the relevant code example

### DIFF
--- a/documentation/tutorials/get-started.md
+++ b/documentation/tutorials/get-started.md
@@ -186,7 +186,7 @@ defmodule MyApp.Accounts.User do
     tokens do
       token_resource MyApp.Accounts.Token
       signing_secret fn _, _ ->
-        Application.fetch_env!(:my_app, :token_signing_secret)
+        Application.fetch_env(:my_app, :token_signing_secret)
       end
     end
   end
@@ -236,7 +236,7 @@ At its simplest you should so something like this:
 # in lib/my_app/accounts/user.ex
 
 signing_secret fn _, _ ->
-  Application.fetch_env!(:my_app, :token_signing_secret)
+  Application.fetch_env(:my_app, :token_signing_secret)
 end
 ```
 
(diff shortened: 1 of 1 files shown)


documentation/tutorials/get-started.md
```diff
@@ -186,7 +186,7 @@ defmodule MyApp.Accounts.User do
     tokens do
       token_resource MyApp.Accounts.Token
       signing_secret fn _, _ ->
-        Application.fetch_env(:my_app, :token_signing_secret)
+        Application.fetch_env!(:my_app, :token_signing_secret)
```
zachdaniel, line 189: This would need to change to `{:ok, Application.fetch_env!(:my_app, :token_signing_secret)}`

I think we should keep this change as is, because `fetch_env` returns `{:ok, value} | :error`, which matches the callback signature required here.

documentation/tutorials/get-started.md
```diff
@@ -220,6 +220,39 @@ identifier.
...
+```
+
+Then, specify the secret token in the config file:
+
+```elixir
+# in config/config.exs
+config :my_app, :token_signing_secret, "some_super_secret_random_value"
+```
+
+> **NOTE:** The signing secret should probably not be committed to source
```
zachdaniel, line 250: Lets use an admonition block here, and remove the word "probably"

```
> ### the signing secret must not be committed to source control {: .warning}
> 
> ....
```

arcanemachine:
OK @zachdaniel I made the changes suggested in both of your comments. Hope it looks good.

zachdaniel (approved):

PR #693, opened 2024-05-19
Update Google OAuth Tutorial

joseph-lozano:
Small doc PR that reflects what worked for me when setting up google oauth.

The base_url is optional, so I removed it from the tutorial.


jimsynz (approved):
thank you

PR #736, opened 2024-06-30
Update confirmation.md to add missing migration steps

msonawane:


zachdaniel:
I think we should omit the output personally. That output will be slightly different depending on the data layer, different circumstances etc. for some data layers no migrations are necessary even :) I'd suggest adding a one liner "don't forget to run `mix ash.codegen` and `mix ash.migrate` if you're using a data layer that uses migrations"

PR #749, opened 2024-07-19
Do a compile-time check to make sure that the configured `token_resource` is an Ash.Resource

simpers:
## Issue
During a refactor of the auth resources, the `tokens.token_resource` field was not updated properly on my user resource, and so I ended up getting a very vague `:error` result when trying to create tokens, without any other context. It was caused by the module not existing, but the current checks only check that the value configured is an atom.

## Errors

Initial error when trying to log in:
```elixir
[error] | file=gen_server.erl line=2474 mfa=:gen_server.error_info/8 [22:44:23.824]: GenServer #PID<0.1271.0> terminating
** (MatchError) no match of right hand side value: :error
    (ash_authentication 4.0.1) lib/ash_authentication/strategies/password/sign_in_preparation.ex:117: AshAuthentication.Strategy.Password.SignInPreparation.generate_token/3
    (ash_authentication 4.0.1) lib/ash_authentication/strategies/password/sign_in_preparation.ex:44: anonymous fn/3 in AshAuthentication.Strategy.Password.SignInPreparation.prepare/3
    (ash 3.2.4) lib/ash/actions/read/read.ex:2097: anonymous fn/2 in Ash.Actions.Read.run_after_action/2
    (elixir 1.17.2) lib/enum.ex:4858: Enumerable.List.reduce/3
    (elixir 1.17.2) lib/enum.ex:2585: Enum.reduce_while/3
    ...
```

After adding some logging, I got this error:
```elixir
[error] | file=lib/ash_authentication/jwt.ex line=119 mfa=AshAuthentication.Jwt.token_for_user/3 [23:26:09.713]: Failed to generate token for user: "Module `Lpe.Resources.Token` is not an Ash resource"
```

## Improvement

Improved compile-time validation of the token_resource option of the tokens DSL by checking that the passed value is an `Ash.Resource`.

## Further improvements?

Could this be improved to test it being an `AshAuthentication.TokenResource` directly?

### DIFF
--- a/lib/ash_authentication/verifier.ex
+++ b/lib/ash_authentication/verifier.ex
@@ -109,23 +109,16 @@ defmodule AshAuthentication.Verifier do
     end
   end
 
-  @spec is_token_resource(any()) :: boolean() | {boolean(), any()}
-  defp is_token_resource(module) when is_atom(module) do
-    with true <- function_exported?(module, :spark_is, 0),
-         Ash.Resource <- module.spark_is() do
-      true
-    else
-      _ ->
-        {false, module}
-    end
+  @spec token_resource?(any()) :: {boolean(), any()}
+  defp token_resource?(module) do
+    {Spark.Dsl.is?(module, Ash.Resource), module}
   end
-  defp is_token_resource(_other), do: false
 
   defp validate_token_resource(dsl_state) do
     if_tokens_enabled(dsl_state, fn dsl_state ->
       with {:ok, resource} when is_truthy(resource) <-
              Info.authentication_tokens_token_resource(dsl_state),
-           true <- is_token_resource(resource) do
+           {true, _resource} <- token_resource?(resource) do
         :ok
       else
         {:ok, falsy} when is_falsy(falsy) ->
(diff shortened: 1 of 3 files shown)


lib/ash_authentication/verifier.ex
```diff
@@ -109,11 +109,23 @@ defmodule AshAuthentication.Verifier do
     end
   end
 
+  @spec is_token_resource(any()) :: boolean() | {boolean(), any()}
+  defp is_token_resource(module) when is_atom(module) do
+    with true <- function_exported?(module, :spark_is, 0),
```
zachdaniel, line 114: I would suggest doing `Spark.Dsl.is?(Ash.Resource)` :)

simpers: That's exactly what I was looking for! Haha thanks

lib/ash_authentication/verifier.ex
```diff
@@ -139,4 +151,18 @@ defmodule AshAuthentication.Verifier do
...
+  # I'm sure this has already been done somewhere else in the Sparkiverse,
+  # so maybe this can be removed and replaced by whatever Spark usually uses
+  # when logging module names?
+  defp bad_token_resource_to_string(module) when is_atom(module) do
+    module
+    |> to_string()
+    |> String.split(".")
+    |> tl()
+    |> Enum.join(".")
+  end
```
simpers, line 164: What about this logic here, @zachdaniel? I know you're logging module names so I assume that you're doing something similar somewhere in Spark?

zachdaniel: oh, right, all ou need here is `inspect(module)`

simpers: > oh, right, all ou need here is `inspect(module)`

Now I feel very silly, indeed!

Fixed, though!

zachdaniel (approved):

PR #760, opened 2024-08-05
improvement: enable custom `http_adapters`

vitormazzi:
Allow users to customize the behavior of the http client used by the oauth-based strategies by configuring a http adapter.


zachdaniel:

PR #766, opened 2024-08-11
fix: allow overriding strategy defaults

zachdaniel:
`auto_set_fields` doesn't set defaults, it sets the values *always* and so things like `authorization_params` were not settable


zachdaniel:
@jimsynz i was having trouble locally with asdf so I updated the .tool-versions file. But now we have what I think are unrelated dialyzer errors. However something strange is happening with dialyzer locally, I think because an explicit PLT file is configured? What is the reason for that? Could you take a look at the failures?

jimsynz:
@zachdaniel dialyzer issues fixed.

PR #770, opened 2024-08-15
fix: actually use 'trusted_audiences' in oauth2 strategy

atoncetti:
Currently `trusted_audiences` is not usable since it is not passed to Assent as a config value.

This pull request makes `trusted_audiences` a secret so it is configurable at runtime and sets `trusted_audiences` in the assent config.


zachdaniel (approved):

zachdaniel:

PR #782, opened 2024-09-06
Add AshAuthentication igniter installer

zachdaniel:
This generates resources & configs, but not plugs/phoenix related code. It also adjusts the documentation to align with the installer/be getting-started oriented.


zachdaniel:
@jimsynz thought you might want to see how this looks. To try it out, you could clone it down and say

```
mix igniter.new my_app --with phx.new --install ash,ash_postgres
# and then (so you can see the diff)
cd my_app

mix igniter.install ash_authentication@../ash_authentication
```

It's uh...pretty awesome :) Seems mega convenient. I'm planning on leaving out the phoenix code *and* the plug code from this. We can tell people in the getting-started guide "if you are using phoenix, run `mix igniter.install ash_authentication_phoenix`, otherwise, copy this controller into your app". So actually...I'll write some tests next week and we can ship it.

zachdaniel:
@jimsynz so credo is annoying me here. It wants me to do aliases, but the ending of the modules in igniter mostly map to modules that exist in Elixir. I.e `Igniter.Code.Module` I could split it up, and do `Code.Module` but that screws up the greppability of these things, I hate splitting modules like that up. I'll do it if you make me, but I personally prefer using the full names in this case, enough to make a stink about it 😆

zachdaniel:
Alright, this isn't quite ready, need to fix a few things first

lib/mix/tasks/ash_authentication.install.ex
```diff
@@ -0,0 +1,308 @@
...
+        end
+        """)
+        |> Ash.Resource.Igniter.add_action(token_resource, """
+        read :get_confirmation_changes do
+          argument :jti, :string, allow_nil?: false, sensitive?: true
+          get? true
+
+          prepare AshAuthentication.TokenResource.GetConfirmationChangesPreparation
+        end
+        """)
```
jimsynz, line 260: these (`*_confirmation_*`) actions are only required if the confirmation add-on has been added.

zachdaniel: Interesting...I don't see them being generated conditionally under the hood. Am I missing something there?

zachdaniel:
I need to ship something in `ash` core for this to be ready, but also I've realized that all of these generated things should be generated with descriptions. Will do that too.

zachdaniel:
Hopefully will be ready to merge tomorrow.

lib/mix/tasks/ash_authentication.install.ex
jimsynz: If they're not that's an oversight. Also doesn't help that they've been essentially invisible until now.

zachdaniel: I think actually it's because you can't look at other resource information in transformers. So you can't know if you need to add them or not.

zachdaniel:
@jimsynz okay, this is ready for final review :) (I hope)

PR #785, opened 2024-09-15
fix: setup options properly for ash 3.0

pshoukry:
The way options where passed tenant was ignored by the Ash.create as it was not translated as options.


zachdaniel (approved):

zachdaniel:

pshoukry:
@zachdaniel most welcome!

PR #795, opened 2024-10-09
improvement: register with magic link

### LINKED ISSUE #554: Add support for registration to magic link strategy
## Context

This originally came from https://elixirforum.com/t/using-magic-links-for-registration/61166. It may also require changes to Ash Authentication Phoenix.

## Problem

The magic link strategy doesn't support registration. It can only be used to sign in existing users/resources.

## Why?

It's not possible to use the magic link strategy as a primary form of auth, you still need another strategy for registration.

## Possible Solution

Look at other strategies for inspiration and add a registration action (and anything needed to support it).

zachdaniel:
registering with magic link turns the read action used to request a magic link into an upsert action.

improvement: support `auto_confirm_actions` on confirmation addon

these actions will automatically apply their confirmation attribute, and will not invoke the sender

improvement: support `ignore_actions` on confirmation addon

these actions will not apply their confirmation attribute, and will not invoke the sender

Closes #554


zachdaniel:
Well. I implemented this as upserts and then I found the issue mentioning this and reading your explanation of how it ought to be done I now know that I did it completely wrong 😆

zachdaniel:
Maybe? I'm actually not sure. The benefits of doing it as an upsert like this are that it can be done atomically. If we do a read, then create a sign up token, we introduce race conditions/potential weird behavior, like having to handle cases where you've requested a token, not used it, signed up using email, in which case it would look like you're trying to sign up for an account that exists.

This way is quite simple by comparison/

zachdaniel:
Alright, now I'm not so sure 😆 @jimsynz lets circle back, we can discuss how this has all gone for me and figure out a way forward :)

zachdaniel:
Alright, just going to close this. It was done incorrectly :D

PR #796, opened 2024-10-09
improvement: support registration via magic link

### LINKED ISSUE #554: Add support for registration to magic link strategy

zachdaniel:
Setting `enable_registration? true` turns the sign in action into a user upsert
instead of a read action. With `enable_registration? true` this also generates
a special token that isn't linked to a user to enable this upsert.

improvement: prevent account takeover hijacking by protecting against upserts against unconfirmed records

This will now produce a StaleRecordError if attempted. More docs have been included about this, in promiment
places. Additionally, you will now get a *compile time error* if you have oauth2 sign ups or magic link alongside
user/pass.

Closes #554


zachdaniel:
I think I'm just going to add in the confirmation upsert protection in this same PR, will let me test a bit more in concert.

zachdaniel:
@jimsynz if I could get a review on this in the next couple days that would be great 🙇 I'll be doing the magic link generators next.

zachdaniel:
Don't hate me, but magic link generators are now a part of this PR also. They got intertwined w/ various fixes/improvements that would have made it duplicative work to review not as a whole unit anyway.

jimsynz (approved):
magic!

PR #800, opened 2024-10-10
Fix OIDC strategy

schutm:
When creating an EntraID (Azure AD) integration I encountered several problems:

1. The DSL wasn't found and resulted in a compilation error, when using code like
```
alias AshAuthentication.Strategy.{Custom, Oidc}
use Custom, entity: Dsl.dsl()
```
2. The assent-library uses strings for the values of ```client_authentication_method``` in stead of atoms
3. Not really a problem but for esthetics the providers was still set to :oauth2 and is now :oidc


jimsynz (approved):
thanks

PR #807, opened 2024-10-21
Fix Apple strategy

miguel-s:
Since the changes in #800 The OIDC strategy takes strings instead of atoms for the `client_authentication_method`.
The Apple dsl was still casting this option to an atom, which prevented compilation with an error.

Removed the casting from atom to string and everything compiles again :)


zachdaniel:

PR #833, opened 2024-11-12
fix(RequestPasswordReset): fails when action called directly.

jimsynz:
In the case where the request password reset action was called directly without the wrapping `Strategy.action/4` the inner `get_by...` action is denied by the default policy.


sevenseacat:
Not even denied by the default policy, it raised an access error didn't it? Either way if thos solves it I am a happy panda

PR #834, opened 2024-11-12
fix(igniter.add_strategy): Generated password reset action names did not match the defaults.

jimsynz:


sevenseacat (approved):
Perfect :)

PR #842, opened 2024-11-17
Update documentation for HashPasswordChange

LannyBose:
When referencing how to make a manual password change, I felt the included documentation could use some sprucing up.

Changes include:
- `Ash.Changeset.new/2` seems to be gone
- Setting the context via `set_context` didn't work as expected when handed to `HashPasswordChange.change`. It kept resulting in `BadMapError`
- Providing explicit direction how to set an argument within the `update` Resource DSL

It's very possible my first chunk of code isn't idiomatic, but it did work. Happy to receive adjusting feedback there!


jimsynz (changes requested):
Hi @LannyBose 👋 

Thanks for the contribution. Unfortunately they're not quite right. If you can update your PR I'll be happy to merge it, otherwise I'll swing back in a day or two and make the changes myself.

jimsynz:
Closing for inactivity. Thanks anyway.

PR #860, opened 2024-12-29
Fixed confirmation docs

brunoripa:
Updated the confirmation docs with the latest findings


zachdaniel:
Thanks for the PR! I needed to make some different clarifications, and there was some incorrect info in the previous docs. I've updated here instead: d0f0384e5d6fc46a64cd79804fc4b6abd42c00b6

PR #861, opened 2024-12-29
Account confirmation

brunoripa:
This PR tries to find a nice way to make the account confirmation mandatory in order to sign in.

### DIFF
--- a/lib/ash_authentication/strategies/password.ex
+++ b/lib/ash_authentication/strategies/password.ex
@@ -94,7 +94,8 @@ defmodule AshAuthentication.Strategy.Password do
   See the [Testing guide](/documentation/topics/testing.md) for tips on testing resources using this strategy.
   """
 
-  defstruct confirmation_required?: false,
+  defstruct require_confirmed_with?: false,
+            confirmation_required?: false,
             hash_provider: AshAuthentication.BcryptProvider,
             hashed_password_field: :hashed_password_field,
             identity_field: :username,
@@ -128,6 +129,7 @@ defmodule AshAuthentication.Strategy.Password do
   use Custom, entity: Dsl.dsl()
 
   @type t :: %Password{
+          require_confirmed_with?: boolean,
           confirmation_required?: boolean,
           hash_provider: module,
           hashed_password_field: atom,
--- a/lib/ash_authentication/strategies/password.ex
+++ b/lib/ash_authentication/strategies/password.ex
@@ -94,7 +94,7 @@ defmodule AshAuthentication.Strategy.Password do
   See the [Testing guide](/documentation/topics/testing.md) for tips on testing resources using this strategy.
   """
 
-  defstruct require_confirmed_with?: nil,
+  defstruct require_confirmed_with: nil,
             confirmation_required?: false,
             hash_provider: AshAuthentication.BcryptProvider,
             hashed_password_field: :hashed_password_field,
@@ -129,7 +129,7 @@ defmodule AshAuthentication.Strategy.Password do
   use Custom, entity: Dsl.dsl()
 
   @type t :: %Password{
-          require_confirmed_with?: :atom | nil,
+          require_confirmed_with: :atom | nil,
           confirmation_required?: boolean,
           hash_provider: module,
           hashed_password_field: atom,
--- a/lib/ash_authentication/strategies/password.ex
+++ b/lib/ash_authentication/strategies/password.ex
@@ -94,7 +94,7 @@ defmodule AshAuthentication.Strategy.Password do
   See the [Testing guide](/documentation/topics/testing.md) for tips on testing resources using this strategy.
   """
 
-  defstruct require_confirmed_with?: false,
+  defstruct require_confirmed_with?: nil,
             confirmation_required?: false,
             hash_provider: AshAuthentication.BcryptProvider,
             hashed_password_field: :hashed_password_field,
@@ -129,7 +129,7 @@ defmodule AshAuthentication.Strategy.Password do
   use Custom, entity: Dsl.dsl()
 
   @type t :: %Password{
-          require_confirmed_with?: boolean,
+          require_confirmed_with?: :atom | nil,
           confirmation_required?: boolean,
           hash_provider: module,
           hashed_password_field: atom,
--- a/lib/ash_authentication/strategies/password/actions.ex
+++ b/lib/ash_authentication/strategies/password/actions.ex
@@ -38,7 +38,25 @@ defmodule AshAuthentication.Strategy.Password.Actions do
     |> Ash.read()
     |> case do
       {:ok, [user]} ->
-        {:ok, user}
+        case strategy.require_confirmed_with? do
+          nil ->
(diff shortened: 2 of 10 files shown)


lib/ash_authentication/strategies/password.ex
```diff
@@ -94,7 +94,8 @@ defmodule AshAuthentication.Strategy.Password do
   See the [Testing guide](/documentation/topics/testing.md) for tips on testing resources using this strategy.
   """
 
-  defstruct confirmation_required?: false,
+  defstruct require_confirmed_with?: false,
+            confirmation_required?: false,
```
zachdaniel, line 98: This is no longer required.

lib/ash_authentication/strategies/password.ex
```diff
@@ -94,7 +94,8 @@ defmodule AshAuthentication.Strategy.Password do
   See the [Testing guide](/documentation/topics/testing.md) for tips on testing resources using this strategy.
   """
 
-  defstruct confirmation_required?: false,
+  defstruct require_confirmed_with?: false,
```
zachdaniel, line 97: The option should not end in a `?`, that is only for boolean options. This should be `nil | atom`

lib/ash_authentication/strategies/password/dsl.ex
```diff
@@ -125,6 +125,12 @@ defmodule AshAuthentication.Strategy.Password.Dsl do
           default: {60, :seconds},
           doc:
             "A lifetime for which a generated sign in token will be valid, if `sign_in_tokens_enabled?`. Unit defaults to `:seconds`."
+        ],
+        require_confirmed_with?: [
+          type: :atom,
+          required: false,
+          doc: "Whether a new account must be confirmed in order to log in.",
+          default: :confirmed_at
```
zachdaniel, line 133: This can't have a default value. It should default to `nil`.

lib/ash_authentication/strategies/password/actions.ex
```diff
@@ -38,7 +38,25 @@ defmodule AshAuthentication.Strategy.Password.Actions do
     |> Ash.read()
     |> case do
       {:ok, [user]} ->
-        {:ok, user}
+        case strategy.require_confirmed_with? do
+          nil ->
+            {:ok, user}
+
+          :confirmed_at ->
```
zachdaniel, line 45: Instead of matching on a literal value, this needs to do something like:

```elixir
case strategy.require_confirmed_with do
  nil ->
    {:ok, user}

  field ->
    if Map.get(user, :confirmed_at) != nil do
      ...
    else
      ...
    end
end
```

Additionally, we need to ensure this attribute is selected in the query, so add something like:

```elixir
|> Query.new()
|> Ash.Query.ensure_selected(List.wrap(strategy.require_confirmed_with))
```

zachdaniel:
This is a great contribution, thanks! Had some comments.

Additionally we need to do this same logic in the `after_action` hook of this module: `AshAuthentication.Strategy.Password.SignInPreparation`, for backwards compatibility.

brunoripa:
Thanks, I will accomodate the comments tomorrow !

lib/ash_authentication/strategies/password.ex
brunoripa: Sorry, just to be sure I understand, you are suggesting that the explicit use of `confirmation_required?` is replaced by `require_confirmed_with` ? Isn't this going to be a breaking change for whatever user code is using the flag ?

brunoripa: Also, the `confirmation_required?` relates to the will to have a `password_confirmation` **field**, if I get this correctly (at least that's what I read in the `Password.md` file which was going to update now.

zachdaniel: Oh sorry. It was a brain fart on my end. That option is whether or not password confirmation is required 😂 ignore me.

brunoripa: Oh, happy I spotted it, I'd have been changing this regardless :P

lib/ash_authentication/strategies/password/actions.ex
```diff
@@ -33,12 +33,31 @@ defmodule AshAuthentication.Strategy.Password.Actions do
...
     |> Ash.read()
     |> case do
       {:ok, [user]} ->
-        {:ok, user}
+        case strategy.require_confirmed_with do
+          nil ->
+            {:ok, user}
+
+          field ->
+            if Map.get(user, field) != nil do
```
brunoripa, line 47: I am not entirely sure about this, probably we should explicitly refer to `:confirmed_at` field ?

brunoripa: Given on the fact that users can have multiple confirmation strategies, this comment of mine is useless and can be ignored.

zachdaniel, line 47: we need a safety check to ensure that the value is not `%Ash.NotLoaded{}`, or `%Ash.ForbiddenField{}`. If it is, we need to return the error case, showing that the user must be confirmed.

lib/ash_authentication/strategies/password/sign_in_preparation.ex
```diff
@@ -46,9 +46,23 @@ defmodule AshAuthentication.Strategy.Password.SignInPreparation do
...
-          token_type = query.context[:token_type] || :user
-
-          {:ok, [maybe_generate_token(token_type, record, strategy)]}
+          if user_confirmed_if_needed(record, strategy) do
+            token_type = query.context[:token_type] || :user
+
+            {:ok, [maybe_generate_token(token_type, record, strategy)]}
+          else
+            {:error,
+             AuthenticationFailed.exception(
```
zachdaniel, line 55: This should add a custom exception, like `AshAuthentication.Errors.CannotConfirmUnconfirmedUser`

lib/ash_authentication/strategies/password/actions.ex
```diff
@@ -279,4 +298,11 @@ defmodule AshAuthentication.Strategy.Password.Actions do
 
   def reset(strategy, _params, _options) when is_struct(strategy, Password),
     do: {:error, NoSuchAction.exception(resource: strategy.resource, action: :reset, type: :read)}
+
+  defp user_confirmed?(user, field) do
+    case Map.get(user, field) do
+      value when value in [nil, %Ash.NotLoaded{}, %Ash.ForbiddenField{}] -> false
```
zachdaniel, line 304: We can’t use `in` here because it does an exact equality filter. So we should match separately on each value.

zachdaniel:
One last comment then it’s good to go!

lib/ash_authentication/strategies/password/actions.ex
brunoripa: Thanks, I'll refactor this. I wanted to write a test for all these cases, but I do not know how to generate an user with those values for `:confirmed_at`. Any existing test I can look at for that ?

zachdaniel (approved):

zachdaniel:

PR #869, opened 2025-01-03
Fix identities concatenation to include primary_key in proper format

jechol:
This PR addresses an issue where identities concatenated with primary_key does not preserve the correct nested structure. Specifically, when:

* primary_key is [:id]
* identities is [[:org, :staff_number], [:ssn], [:email]]

The resulting value becomes [[:org, :staff_number], [:ssn], [:email], :id], but it should be [[:org, :staff_number], [:ssn], [:email], [:id]].

This fix ensures that primary_key is correctly wrapped in a list when concatenated, resulting in the expected structure.


zachdaniel (approved):

zachdaniel:

PR #870, opened 2025-01-04
UID conversion to string added

brunoripa:
Some providers like `Github` return an `uid` value as an _integer_, not `:string` which is the only type that was accepted. Now the value gets converted if it's not a string.

### DIFF
--- a/lib/ash_authentication/user_identity/upsert_identity_change.ex
+++ b/lib/ash_authentication/user_identity/upsert_identity_change.ex
@@ -38,6 +38,7 @@ defmodule AshAuthentication.UserIdentity.UpsertIdentityChange do
       |> Map.values()
       |> Enum.reject(&is_nil/1)
       |> List.first()
+      |> convert()
 
     changeset
     |> Changeset.change_attribute(cfg.user_id_attribute_name, user_id)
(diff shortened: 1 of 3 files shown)


lib/ash_authentication/user_identity/upsert_identity_change.ex
```diff
@@ -38,6 +38,7 @@ defmodule AshAuthentication.UserIdentity.UpsertIdentityChange do
       |> Map.values()
       |> Enum.reject(&is_nil/1)
       |> List.first()
+      |> convert()
```
zachdaniel, line 41: I think we can just to `to_string()` here since that doesn’t do anything if it is already a string.

test/ash_authentication/strategies/oauth2/actions_test.exs
```diff
@@ -42,6 +42,39 @@ defmodule AshAuthentication.Strategy.OAuth2.ActionsTest do
...
+               )
+
+      assert signed_in_user.id == user.id
+
+      assert {:ok, _claims} =
+               Jwt.peek(signed_in_user.__metadata__.token)
+
+      # I need to manipulate the token generation to make this work
+      # Is there a proper way instead of mocking stuff around ?
+      # assert claims["sub"] =~ 1234
```
zachdaniel, line 75: I don’t think we can reasonably do this. We store the value as a string and people will need to handle that whenever they use the value.

PR #871, opened 2025-01-05
Allow Custom Attribute Name for `created_at`

mmaa:
By default, Phoenix uses `inserted_at` for the column name in `timestamps()`. This allows the customization of the `created_at` attribute name to align with Phoenix defaults.

### DIFF
--- a/lib/ash_authentication/token_resource.ex
+++ b/lib/ash_authentication/token_resource.ex
@@ -14,6 +14,11 @@ defmodule AshAuthentication.TokenResource do
           The Ash domain to use to access this resource.
           """
         ],
+        created_at_attribute_name: [
+          type: :atom,
+          doc: "The name of the `created_at` attribute on this resource.",
+          default: :created_at
+        ],
         expunge_expired_action_name: [
           type: :atom,
           doc: """
--- a/lib/ash_authentication/token_resource/transformer.ex
+++ b/lib/ash_authentication/token_resource/transformer.ex
@@ -45,6 +45,8 @@ defmodule AshAuthentication.TokenResource.Transformer do
              public?: true
            ),
          :ok <- validate_jti_field(dsl_state),
+         {:ok, created_at} <-
+           TokenResource.Info.token_created_at_attribute_name(dsl_state),
          {:ok, dsl_state} <-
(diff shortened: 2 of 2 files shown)


zachdaniel (commented):
You should be able to define a test for this purpose. use the token extension in a resource inside that test. Then you can assert on which attributes it has using `Ash.Resource.Info.attributes(TokenResource)`.

```elixir
test "uses created_at" do
  defmodule Token do
    use Ash.Resource, extensions: [....]
  end

  assert Ash.Resource.Info.attribute(Token, :inserted_at)
end
```

mmaa:
Thanks, @zachdaniel. I've added a test which confirms that this change is actually not working as expected. Looks like I'm missing something to make this work.

Using `create_timestamp :inserted_at` does give me the attribute `:inserted_at`, but I still end up with a `:created_at`:
```elixir
> Ash.Resource.Info.attribute_names(Example.TokenWithCustomCreateTimestamp)
MapSet.new([:inserted_at, :extra_data, :purpose, :created_at, :jti, :expires_at, :subject, :updated_at])
```

zachdaniel:
Hm...that's pretty strange. Is the `timestamps()` constructor being used somewhere?

mmaa:
The only references to `timestamps` are in the install task and test.

zachdaniel:
Interesting. I'd have to investigate further but the only other place it could be coming from is the token extensions transformer. Could thing else be adding that created at?

mmaa:
I've updated the test to verify that the custom attribute is present, and that the default one is not. The test is failing. The expectation is that the custom attribute replaces the default one.

zachdaniel:
Checked out your branch, I think you're just missing using your new config:

```elixir
  token do
    created_at_attribute_name :inserted_at
  end
```

in your test resource.

zachdaniel:
Is this ready to go now?

mmaa:
Yes, I think it is. I thought `create_timestamp :inserted_at` was what I needed, but I must not understand what that does.

zachdaniel:
That creates a timestamp called `inserted_at`, but the new option you added still has to be used to instruct `ash_authentication` not to create its default `created_at` attribute.

PR #905, opened 2025-02-11
fix: Ensure that installer generated token revocation checking action is correct.

jimsynz:
Unfortunately, when the igniter installer was introduced it generated an action which did not correctly check for token revocation.

This has been fixed, along with a verifier warning and an update which fixes the issue.

### DIFF
--- a/lib/ash_authentication/token_resource/verifier.ex
+++ b/lib/ash_authentication/token_resource/verifier.ex
@@ -3,40 +3,79 @@ defmodule AshAuthentication.TokenResource.Verifier do
   The token resource verifier.
   """
 
-  use Spark.Dsl.Transformer
+  use Spark.Dsl.Verifier
   require Ash.Expr
-  alias Spark.{Dsl.Transformer, Error.DslError}
+  require Logger
+  alias Spark.{Dsl.Verifier, Error.DslError}
   import AshAuthentication.Utils
+  import AshAuthentication.Validations.Action
 
   @doc false
   @impl true
-  @spec after?(any) :: boolean()
-  def after?(_), do: true
+  @spec verify(map) :: :ok | {:error, term}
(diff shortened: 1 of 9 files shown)


lib/mix/tasks/ash_authentication.install.ex
```diff
@@ -344,8 +344,8 @@ if Code.ensure_loaded?(Igniter) do
           |> Ash.Resource.Igniter.add_action(token_resource, """
           action :revoked? do
```
zachdaniel, line 345:
```suggestion
          action :revoked?, :boolean do
```

lib/ash_authentication/token_resource/verifier.ex
```diff
@@ -3,40 +3,69 @@ defmodule AshAuthentication.TokenResource.Verifier do
...
+        This was an error in our igniter installer previous to version 4.4.9, which allowed revoked tokens to be reused.
+
+        To fix this, run the following command in your shell:
+
+            mix ash_authentication.upgrade 4.4.8 4.4.9
+
+        Or:
+
+          - remove `allow_nil?: false` from these action arguments, and
+          - add `returns :boolean` to the action.
```
zachdaniel, line 56: the `returns` is a second arg to the action

jimsynz: that makes it so much harder to find the action with igniter :sob:

zachdaniel: Fun fact, I guess it can actually be *both* the second arg *or* `returns :boolean` 😅

zachdaniel: It shouldn't make it much harder to find though, because the function finding tools accept a list of possible rarities.

PR #907, opened 2025-02-12
feat: Add a `log_out_everywhere` add-on

jimsynz:
Adds the `log_out_everywhere` add-on which adds a `log_out_everywhere` action to the authentication resource which revokes all tokens for a user when called.


lib/ash_authentication/add_ons/log_out_everywhere/action.ex
```diff
@@ -0,0 +1,80 @@
...
+  end
+
+  defp revoke_all_tokens_for_subject(subject, strategy) do
+    with {:ok, token_resource} <- Info.authentication_tokens_token_resource(strategy.resource),
+         {:ok, tokens} <- get_all_tokens_for_subject(token_resource, subject) do
+      tokens
+      |> Stream.map(
+        &%{
+          jti: &1.jti,
+          purpose: "revocation",
```
sevenseacat, line 54: We explicitly add new tokens to revoke the old ones, we wouldn't simply delete all the existing tokens? That's what I was expecting to see here

sevenseacat (commented):
I can't comment on the meat of the code, but :100: for the add-on!

lib/ash_authentication/add_ons/log_out_everywhere/action.ex
zachdaniel: Good point. Seems like we should only be revoking specific kinds of tokens? Or maybe the purposes revoked should be configurable and should default to the ones we care about like sign_in, magic_link, password_reset etc?

zachdaniel: As for revoking old tokens, I also think that would be a better default strategy, but the default setup does not `require_token_presence_for_authentication`. @jimsynz perhaps we should modify token revocation to just delete tokens when `require_token_presence_for_authentication`? Then we also don't need to check for revocation tokens if that setting is true also. Users can still opt out of the requirement to store all tokens, but when they opt in, AA will then make fewer queries and behave a bit simpler (the way @sevenseacat originally expected it to).

jimsynz: we upsert them which really since we queried for them in the first place is just an update. maybe I should just make it an update.

zachdaniel: From what we were seeing, we do not upsert them. There is not an identity on the jit.

jimsynz: The JTI is the primary key

zachdaniel: Oh, where are we configuring it to be an upsert when revoking?

zachdaniel: aw damnit...its the generators again, they are adding a uuid primary key 😢.

So we need another warning and fix to that. Its fine because current logic checks for a revocation token 😂

zachdaniel: Fix is here, and users will need to follow the instructions when upgrading:

https://github.com/team-alembic/ash_authentication/pull/908

zachdaniel: Alright, so once that PR is merged, we can merge this one, and then we can, as a follow up, remove any place that we are looking up revocation tokens if `require_token_presence_for_authentication?` is set to `true`. Finally, we will want to modify the installers to make all of this the default behavior. We can do that in a follow up to this PR.

zachdaniel: Well, not in that order. We should modify the installers first, because the avoiding a lookup thing is just a perf improvement.

zachdaniel: Okay, this now does a bulk update.

jimsynz:
Okay, so I did a small tidy up but I don't think we should call the action `revoke_all_tokens`, because it's an update it's definitely not "all tokens" unless run in bulk in combination with a query. It's not even really a revoke because it only works on tokens that are already stored, so won't work unless `store_all_tokens?` is enabled.

zachdaniel:
Let's call it `log_out_everywhere` then and only add it when that add_on is used. Then it's very clear what it's for :smile:

PR #908, opened 2025-02-12
fix: ensure that the token resource has only `:jti` as a primary key

zachdaniel:
for users who have used the generators to create their application, they will see a compile time error explaining how to fix this issue.

CleanShot 2025-02-12 at 17 22 29@2x

EDIT:

slightly fixed wording:

           You are likely seeing this as a by-product of an error with the generators 
           that added a `uuid_primary_key :id` to the token resource.

### DIFF
--- a/lib/ash_authentication/token_resource/transformer.ex
+++ b/lib/ash_authentication/token_resource/transformer.ex
@@ -156,8 +156,70 @@ defmodule AshAuthentication.TokenResource.Transformer do
              get_token_action_name,
              &build_get_token_action(&1, get_token_action_name)
            ),
-         :ok <- validate_get_token_action(dsl_state, get_token_action_name) do
+         :ok <- validate_get_token_action(dsl_state, get_token_action_name),
+         :ok <- validate_primary_key(dsl_state) do
       {:ok, dsl_state}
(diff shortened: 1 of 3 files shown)


lib/ash_authentication/token_resource/transformer.ex
```diff
@@ -156,11 +156,73 @@ defmodule AshAuthentication.TokenResource.Transformer do
...
+         :ok <- validate_primary_key(dsl_state) do
       {:ok, dsl_state}
     end
   end
 
+  defp validate_primary_key(dsl_state) do
+    dsl_state
+    |> Ash.Resource.Info.attributes()
+    |> Enum.filter(& &1.primary_key?)
+    |> Enum.map(& &1.name)
```
jimsynz, line 169:
```suggestion
    |> Ash.Resource.Info.primary_key()
```

jimsynz (approved):
clean up

lib/ash_authentication/token_resource/transformer.ex
zachdaniel: We can't use this yet because this transformer runs before the id transformer.

PR #914, opened 2025-02-18
fix(JWT): Generate tenant claims and validate them.

jimsynz:
I don't like that it always places a `"tenant" => nil` claim in the token.


zachdaniel:
Yeah we need to find way to not have tenant set to nil. We also need to be sure that this method isn't requiring that all non multi tenant tokens explicitly have that key and is set to nil because that will log out all users for all non multi tenant usages of AA on upgrade.

lib/ash_authentication/jwt/config.ex
```diff
@@ -77,6 +82,51 @@ defmodule AshAuthentication.Jwt.Config do
...
+      case Keyword.fetch(options, :tenant) do
+        {:ok, ^tenant} -> true
+        _ -> false
+      end
+    else
+      true
+    end
+  end
+
+  defp multitenancy_required?(user_resource) do
```
zachdaniel, line 119: It's not about if the token is multi tenant, but if the user is being multi tenant. When we generate the token we need to set the tenant into it, but only if the tenant option is set when generating the token. You can have resources that support being accessed both with a tenant and without.

zachdaniel: Actually I think we shouldn't worry about any of it. If the tenant option is set (to anything other than nil) in this code, then we write the tenant into the claims. If there is a tenant claim when validating, or if the tenant option is set, then we validate that the two match.

zachdaniel: Nope, I'm wrong. We do want to check if the user (but not the token) supports multitenancy, and if so then we add and validate the tenant. That might make it easier because in *those* cases it actually makes sense to add a nil tenant claim if it is nil.

jimsynz: Actually I think I just realised that we can just make it optional whether we add it to the config if required.

lib/ash_authentication/jwt/config.ex
```diff
@@ -48,6 +48,18 @@ defmodule AshAuthentication.Jwt.Config do
       &Joken.generate_jti/0,
       &validate_jti(&1, &2, &3, opts)
     )
+    |> maybe_add_tenant_claim(opts[:tenant])
+  end
+
+  defp maybe_add_tenant_claim(cfg, nil), do: cfg
+
+  defp maybe_add_tenant_claim(cfg, tenant) do
+    Config.add_claim(
```
zachdaniel, line 57: So the main thing missing here is that if there is a tenant claim, and no `opts[:tenant]` then it is invalid. How to we add this?

zachdaniel: Alright,  I remember the simpler answer. its "does the user resource support multi tenancy", and if so there is a tenant claim that must match, nil or otherwise.

PR #915, opened 2025-02-18
Add multi tenant tests

DGollings:
Good morning, quick draft PR
https://github.com/team-alembic/ash_authentication/pull/915

I:
extended Example.Repo with a multi tenant enabled example tables
(originally had a separate Repo but that seemed to complicate more than help)
with either suffix `MultiTenant` or prefix `mt_`
removed GraphQL and OpenAPI stuff, have not worked with them and don't think its necessary (yet?) in this case
added a resource `Organisation` which everything else `belongs_to`

BUG (?): `mix test` breaks on various api/plug tests, and I don't really understand why. Although I would have to add I did not try running them before my changes (but my changes should be unrelated)

bonus: added devenv.sh so I could do a simple `devenv up` and `mix test`, we can remove that of course

how is this for an initial direction?


test/support/example_multi_tenant.ex
```diff
@@ -0,0 +1,13 @@
...
+
+  resources do
+    resource ExampleMultiTenant.Organisation
+    resource ExampleMultiTenant.User
+    resource ExampleMultiTenant.UserWithTokenRequired
+    resource ExampleMultiTenant.Token
+    resource ExampleMultiTenant.UserIdentity
+    resource ExampleMultiTenant.UserWithRegisterMagicLink
+  end
+end
```
alexslade, line 13: (I'm not an Ash maintainer, just another dev with Ash + multitenancy trying to help, so please prioritise any comments from the Ash team over mine) 

It feels odd to me to multi-tenant tokens? The first step for many apps is to authenticate, and _that_ informs which tenant is used from there on out. So the tenancy doesn't exist before the auth itself. 

I assume you're setting tenancy with some other method like a subdomain? 

(Possibly relevant discord link: https://discord.com/channels/711271361523351632/1253437793053577226/1341155047060537395) 

> I'm not sure why you'd bother making the token resource multi tenant tbh

zachdaniel: Multitenancy may also come from something like a subdomain or a header, so there is a case for it.

DGollings: I agreed with someone else's comment that its weird to add MT to a token, but also that it should be possible, so left it in :)

zachdaniel:
These changes look generally good to me, but  I think we should take out the dev env stuff for now. We can talk about adding that in a separate PR.

zachdaniel:
I actually have a use case for these tenant resources in the short term, so I'd like to merge them as soon as the devenv stuff is merged. We want to add the tenant to the token and I need to test that behavior.

DGollings:
reverted devenv and marked as ready to review/merge

zachdaniel:

PR #921, opened 2025-02-21
fix: ensure token expunger sets context that allows it to run

zachdaniel:
improvement: authorize with error to make expunge errors clearer
test: add policies to tokens in test to detect issues later


jimsynz (approved):
got dammit

PR #928, opened 2025-03-02
chore: Add failing test for atomic updates that are accidentally triggering the `log_out_everywhere` functionality

sevenseacat:
See conversation in Discord from [here](https://discord.com/channels/711271361523351632/1253437793053577226/1345745085060612107) onwards.

A non-atomic update that doesn't touch the `hashed_password` field appears to be triggering the `log_out_everywhere` functionality. The auth token on the user is no longer valid afterwards.

The non-atomic `update` doesn't run any queries to update tokens, but the atomic `update` does:

```sql
UPDATE "tokens" AS t0 SET "purpose" = $1, "updated_at" = (CASE WHEN $2::text::text != 
t0."purpose"::text::text THEN $3::timestamp ELSE t0."updated_at"::timestamp END)::timestamp 
WHERE (NOT (t0."purpose"::text::text = ANY($4::text[]::text[]))) AND (t0."subject"::text::text
 = $5::text::text) ["revocation", "revocation", ~U[2025-03-02 14:06:05.209533Z], 
["revocation"], "user?id=37ad93c0-9332-4a37-a0f9-141ac33db506"]
```


zachdaniel:
Fixed in Ash and updated here.

PR #945, opened 2025-03-22
fix: Multitenant magic link redirect

ken-kost:
Related to the issue in: https://github.com/alemoreau/ash_multitenancy_repro

Not sure if right approach but it does fix the issue.

### DIFF
--- a/lib/ash_authentication/strategies/magic_link/request.ex
+++ b/lib/ash_authentication/strategies/magic_link/request.ex
@@ -9,7 +9,7 @@ defmodule AshAuthentication.Strategy.MagicLink.Request do
 
   @doc false
   @impl true
-  def run(input, _opts, context) do
+  def run(input, opts, context) do
     strategy = Info.strategy_for_action!(input.resource, input.action.name)
 
     identity = ActionInput.get_argument(input, strategy.identity_field)
@@ -44,7 +44,7 @@ defmodule AshAuthentication.Strategy.MagicLink.Request do
 
       {:ok, user} ->
         with {sender, send_opts} <- strategy.sender,
-             {:ok, token} <- MagicLink.request_token_for_identity(strategy, identity, context) do
+             {:ok, token} <- MagicLink.request_token_for(strategy, user, opts, context) do
           sender.send(user, token, Keyword.put(send_opts, :tenant, context.tenant))
 
           :ok
--- a/lib/ash_authentication/strategies/magic_link/request.ex
+++ b/lib/ash_authentication/strategies/magic_link/request.ex
@@ -9,10 +9,10 @@ defmodule AshAuthentication.Strategy.MagicLink.Request do
 
   @doc false
   @impl true
-  def run(input, opts, context) do
+  def run(input, _opts, context) do
     strategy = Info.strategy_for_action!(input.resource, input.action.name)
-
     identity = ActionInput.get_argument(input, strategy.identity_field)
+    context_opts = Ash.Context.to_opts(context)
 
     input.resource
     |> Query.set_context(%{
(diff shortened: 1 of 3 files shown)


lib/ash_authentication/strategies/magic_link/request.ex
```diff
@@ -44,7 +44,7 @@ defmodule AshAuthentication.Strategy.MagicLink.Request do
 
       {:ok, user} ->
         with {sender, send_opts} <- strategy.sender,
-             {:ok, token} <- MagicLink.request_token_for_identity(strategy, identity, context) do
+             {:ok, token} <- MagicLink.request_token_for(strategy, user, opts, context) do
```
zachdaniel, line 47: Lets put this back to the `request_token_for_identity`, but add an `opts` to that, which uses `Keyword.merge` like `request_token_for_identity` does 🙇

lib/ash_authentication/strategies/magic_link/sign_in_preparation.ex
```diff
@@ -49,7 +49,7 @@ defmodule AshAuthentication.Strategy.MagicLink.SignInPreparation do
         query, [record] ->
           if strategy.single_use_token? do
             token_resource = Info.authentication_tokens_token_resource!(query.resource)
-            :ok = TokenResource.revoke(token_resource, token)
+            :ok = TokenResource.revoke(token_resource, token, tenant: context.tenant)
```
zachdaniel, line 52:
```suggestion
            :ok = TokenResource.revoke(token_resource, token, Ash.Context.to_ops(context))
```

lib/ash_authentication/strategies/magic_link/request.ex
zachdaniel: And instead of using the `opts` from `run`, we need to pass in `Ash.Context.to_opts(context)` to that new `opts` argument. Make sure the new `opts` argument is optional for backwards compatibility, i.e `opts \\ []`.

ken-kost: Alright, but regarding backwards compatibility this would mean `opts \\ []` would have to be at the end of the function, as last argument. While for `request_token_for` it's second to last. 

I tried doing it as `request_token_for` and changed everywhere where it's called but now I'm thinking this would be a breaking change for anyone using context as third argument.

That's why I only is putting it at the end, last argument; but that it's inconsistent with the first function so, that's but bugs me. 🐛

lib/ash_authentication/strategies/magic_link/request.ex
```diff
@@ -33,7 +33,8 @@ defmodule AshAuthentication.Strategy.MagicLink.Request do
       {:ok, nil} ->
         with true <- strategy.registration_enabled?,
              {sender, send_opts} <- strategy.sender,
-             {:ok, token} <- MagicLink.request_token_for_identity(strategy, identity, context) do
+             {:ok, token} <-
+               MagicLink.request_token_for_identity(strategy, identity, context, context_opts) do
```
ken-kost, line 37: Wait, not sure should be I passing it here as well.? 🤔

lib/ash_authentication/strategies/magic_link/request.ex
zachdaniel: It's fine if it's inconsistent. It is what it is 😆

lib/ash_authentication/strategies/magic_link/request.ex
zachdaniel: Yes this looks right to me.

zachdaniel:
LMK if the example app works after applying your latest changes :)

ken-kost:
> LMK if the example app works after applying your latest changes :)

image
😎 @zachdaniel

lib/ash_authentication/strategies/magic_link/request.ex
ken-kost: by the power of consistency - it shall pass~!

zachdaniel:

PR #946, opened 2025-03-22
Document creating a helper for testing authenticated LiveViews

sodapopcan:
Howdy!

As per our brief [forum discussion](https://elixirforum.com/t/how-to-test-live-views-with-ash-authentication-plugs/59814/17?u=sodapopcan), here is some documentation on how to create a `register_and_log_in_user` test helper.

### DIFF
--- a/documentation/topics/testing.md
+++ b/documentation/topics/testing.md
@@ -12,3 +12,78 @@ AshAuthentication uses `bcrypt_elixir` for hashing passwords for secure storage,
 # Do NOT set this value for production
 config :bcrypt_elixir, log_rounds: 1
 ```
+
+## Testing authenticated LiveViews
+
+In order to test authenticated LiveViews, you will need to seed a test user and
+log in it.  While you may certainly use a helper that logs in through the UI
+each time, it's a little more efficient to call the sign-in code directly.
+
+This can be done by adding a helper function called `MyAppWeb.ConnCase` found in
+`test/support/conn_case.ex`.  In this example it's called
+`register_and_log_in_user`.
+
+```elixir
+defmodule MyAppWeb.ConnCase do
+  use ExUnit.CaseTemplate
+
+  using do
+    # ...
+  end
+
+  def register_and_log_in_user(%{conn: conn}) do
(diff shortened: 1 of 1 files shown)


sodapopcan:
hrm, this is actually not working (my local tests were flawed).

I put the default user in `seeds.exs` so it's not seeded over and over and that works, but I've done something wrong here.  Will try and figure it out.

sodapopcan:
Ok, I've properly tested it now and it works, I just had a few stupid typos.  I thought I had fully copy/pasted to test... maybe I didn't save?  I dunno, it's all good now.

zachdaniel:
Oh, interesting, I guess you have to do the sign in action still to produce the token metadata, that makes sense.

zachdaniel:
It does make sense for us to have some kind of helper for this at some point, but this documentation is great and will be very helpful in the meantime! Its hard to create a general purpose tool for this given all the strategies/constraints that may be at play etc.

sodapopcan:
Do you see that as a generated helper like phx_new or provided by the package?  Providing it from the package could be a bit of a rabbit hole, I'm not sure.  I'm still understanding the internals and if it would be possible to provide a generic way to sign in a user regardless of strategy.  `sign_in` does quite a bit in many cases though it seems largely the end result is whether or not it's token-based and whether or not `:ash_authenticated?` is set (I may well be missing something, of course).

Otherwise, I was thinking you could provide `_with_<strategy>` suffixed helpers, so like `sign_in_with_password`/`sign_in_with_magic_link`/`sign_in_with_oath` etc.  As well it could have two seed helpers: `register_user` and `register_confimed_user`.  Then they could be used like so:

```
setup [:register_confirmed_user, :sign_in_with_password]
```

Have you all had any thoughts on this?

I'm happy to take a stab at it.

zachdaniel:
Thinking about it, we could likely have a generic helper that generates and stores a token for a test user regardless of strategy using the token resource directly.

sodapopcan:
That would definitely be ideal!  I haven't looked into since posting, but if I do I'll open a PR or an issue.

PR #947, opened 2025-03-23
make sure to point to right function for deprecation for AshAuthentication.Secret.secret_for/3

allenwyma:
Feel free to dismiss the `tool-versions` update, i just like to use the latest otp and elixir when possible.


zachdaniel:

PR #949, opened 2025-03-24
fix: make request_for_token_identity backwards compatible

barnabasJ:


lib/ash_authentication/strategies/magic_link.ex
```diff
@@ -175,7 +175,7 @@ defmodule AshAuthentication.Strategy.MagicLink do
 
   Used by `AshAuthentication.Strategy.MagicLink.RequestPreparation`.
   """
-  def request_token_for_identity(strategy, identity, context, opts \\ [])
+  def request_token_for_identity(strategy, identity, context \\ %{}, opts \\ [])
```
barnabasJ, line 178: As the context was just added with the last version and the options aren't released yet, it might be worth switching them around, to be more consistent with the other functions. WDYT?

zachdaniel: Hmm...okay, let's switch them around. I was avoiding it but its unlikely to cause problems.

zachdaniel:

PR #950, opened 2025-03-26
improvement: delete_session on unsuccessful auth

zachdaniel:
This may look like a security-related change, but it is not, any relevant tokens are always checked for authentication. This is in preparation for a tool that will peek the token and use `assign_new` without reauthenticating, for a better nested liveview experience


jimsynz (approved):
fine

PR #951, opened 2025-03-27
improvement: add `AshAuthentication.Plug.Helpers.assign_new_resources/4`

zachdaniel:
This helper will be used for a new tool in the `LiveUserAuth` module that assigns a new user from session. The only thing it bypasses is the check for `require_token_presence_for_authentication?` because that work will have already been done by the parent LiveView, and the utility is for nested live views.

### DIFF
--- a/lib/ash_authentication/plug/helpers.ex
+++ b/lib/ash_authentication/plug/helpers.ex
@@ -132,7 +132,7 @@ defmodule AshAuthentication.Plug.Helpers do
   @doc """
   Assigns all subjects from their equivalent sessions, if they are not already assigned.
 
-  This is meant to used bvia `AshAuthenticationPhoenix` for nested liveviews.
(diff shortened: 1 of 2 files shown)


lib/ash_authentication/plug/helpers.ex
```diff
@@ -129,6 +129,64 @@ defmodule AshAuthentication.Plug.Helpers do
     end)
   end
 
+  @doc """
+  Assigns all subjects from their equivalent sessions, if they are not already assigned.
+
+  This is meant to used bvia `AshAuthenticationPhoenix` for nested liveviews.
```
jimsynz, line 135:
```suggestion
  This is meant to used via `AshAuthenticationPhoenix` for nested liveviews.
```

jimsynz (approved):
gotta go fast!

PR #961, opened 2025-04-04
Update bcrypt_provider.ex

lifeiscontent:
fix: add guard clause when input is nil

PR #962, opened 2025-04-04
fix: Potential timing attack introduced by #961.

jimsynz:


zachdaniel:
🥶 My bad.

lifeiscontent:
and this is why I am hesitant to do auth stuff 😅  thank you @jimsynz !

jimsynz:
I think it's pretty unlikely to ever be hit but better safe than sorry ❤️

PR #967, opened 2025-04-14
Minor tweaks to Auth0 strategy

agonzalezro:
If you start a project from scratch using the Ash bootstrapper you might end up in a situation where your `User` resource doesn't have an email attribute. This commit explains what to do in that case.

Nit: I've also added an arg to `secret_for` as the previous `/3` is deprecated.


zachdaniel (approved):

zachdaniel:

PR #976, opened 2025-04-30
fix(MagicLink): Add `require_interaction?` option to magic link strategy.

jimsynz:
This is to stop magic link tokens being automatically consumed by email clients, virus scanners, etc.

### DIFF
--- a/documentation/dsls/DSL-AshAuthentication.Strategy.MagicLink.md
+++ b/documentation/dsls/DSL-AshAuthentication.Strategy.MagicLink.md
@@ -79,6 +79,20 @@ Signing in using a magic link token:
     ...> signed_in_user.id == user
     true
 
+## Usage with AshAuthenticationPhoenix
+
+If you are using `AshAuthenticationPhoenix`, and have `require_authentication?` set to `true`, which you very much should, then you will need to add a `magic_sign_in_route` to your router. This is placed in the same location as `auth_routes`, and should be provided the user and the strategy name. For example:
+
+```elixir
+# Remove this if you do not want to use the magic link strategy
+magic_sign_in_route(
+  MyApp.Accounts.User,
+  :sign_in,
+  auth_routes_prefix: "/auth",
+  overrides: [MyApp.AuthOverrides, AshAuthentication.Phoenix.Overrides.Default]
+)
+```
+
 ## Plugs
 
 The magic link strategy provides plug endpoints for both request and sign-in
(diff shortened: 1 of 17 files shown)


documentation/dsls/DSL-AshAuthentication.Strategy.MagicLink.md
```diff
@@ -129,6 +143,7 @@ Strategy for authenticating using local users with a magic link
...
 | [`token_lifetime`](#authentication-strategies-magic_link-token_lifetime){: #authentication-strategies-magic_link-token_lifetime } | `pos_integer \| {pos_integer, :days \| :hours \| :minutes \| :seconds}` | `{10, :minutes}` | How long the sign in token is valid.  If no unit is provided, then `minutes` is assumed. |
 | [`prevent_hijacking?`](#authentication-strategies-magic_link-prevent_hijacking?){: #authentication-strategies-magic_link-prevent_hijacking? } | `boolean` | `true` | Requires a confirmation add_on to be present if the password strategy is used with the same identity_field. |
+| [`require_interaction?`](#authentication-strategies-magic_link-require_interaction?){: #authentication-strategies-magic_link-require_interaction? } | `boolean` | `false` | Whether or not to require user interaction to sign in. If true, the magic link URLs are changed to a `POST` request, and AshAUthenticationPhoenix will show a button to confirm when the page is visited |
```
RoyalIcing, line 146: Typo
```suggestion
| [`require_interaction?`](#authentication-strategies-magic_link-require_interaction?){: #authentication-strategies-magic_link-require_interaction? } | `boolean` | `false` | Whether or not to require user interaction to sign in. If true, the magic link URLs are changed to a `POST` request, and AshAuthenticationPhoenix will show a button to confirm when the page is visited |
```

lib/ash_authentication/strategies/magic_link/dsl.ex
```diff
@@ -43,6 +43,12 @@ defmodule AshAuthentication.Strategy.MagicLink.Dsl do
           doc:
             "Requires a confirmation add_on to be present if the password strategy is used with the same identity_field."
         ],
+        require_interaction?: [
+          type: :boolean,
+          default: false,
+          doc:
+            "Whether or not to require user interaction to sign in. If true, the magic link URLs are changed to a `POST` request, and AshAUthenticationPhoenix will show a button to confirm when the page is visited"
```
RoyalIcing, line 50: Typo
```suggestion
            "Whether or not to require user interaction to sign in. If true, the magic link URLs are changed to a `POST` request, and AshAuthenticationPhoenix will show a button to confirm when the page is visited"
```

zachdaniel:
LGTM but we definitely need to update the add strategy task to include this by default

jimsynz:
@zachdaniel please review again in the morning - I've made some changes that I'd like a second pair of eyeballs on and I intend to augment the confirmation in the same way tomorrow.

lib/ash_authentication/strategies/magic_link/transformer.ex
```diff
@@ -171,4 +173,25 @@ defmodule AshAuthentication.Strategy.MagicLink.Transformer do
...
+    bypassing_error? =
+      :ash_authentication
+      |> Application.get_env(:bypass_require_interaction_for_magic_link?, false)
+
+    if bypassing_error? do
+      :ok
+    else
+      Logger.warning(fn ->
+        """
+        `require_interaction?` must be set to true on the #{inspect(strategy.name)} magic link strategy for #{inspect(strategy.resource)}.
```
zachdaniel, line 189:
```suggestion
        `require_interaction?` should be set to true on the #{inspect(strategy.name)} magic link strategy for #{inspect(strategy.resource)}.
```

lib/mix/tasks/ash_authentication.add_strategy.ex
```diff
@@ -426,6 +427,7 @@ if Code.ensure_loaded?(Igniter) do
           end
 
           defp body(params) do
+            # TODO: Set to your magic link acceptance URL.
             url = url(~p"/auth/user/magic_link/?token=\#{params[:token]}")
```
zachdaniel, line 431: Will this line up with the default? We shouldn't have a TODO that the user has to do post-install hidden in a file. We also changed the URL for `confirmation`, I think we should change it in the same way for this, so that it is something like `/magic_link`. So we should update this to match that suggestion. In the warning, we should tell users to update the URL to be what the old default is in order to upgrade, i.e `path: "/auth/user/magic_link", token_as_route_param?: false`, or link to a guide that explains what to do, because just changing `require_interaction?` to `true` is not enough (they also have to modify their router, and make these changes etc.

jimsynz: it will work with the default plug-based URL. it should be changed if they're using the new route in AAP though.

zachdaniel: Right, so I think we should default to the one that works w/ Phoenix since that is 99% of installations.

jimsynz: turns out we have this awesome variable in scope `use_web_module` :)

zachdaniel (approved):

barnabasJ:
> This might be a dumb question, so sorry in advance.
> 
> Running phoenix with "magic link" strategy and "requires interaction" fails signin/register at interaction page due to missing CSRF token, shouldn't `get_csrf_token()` be included in lib/ash_authentication/strategies/magic_link/sign_in_form.html.eex ? although I know this package is ash_authentication and not ash_authentication_phoenix so I'm really just trying to understand how to go about fixing this problem.

IIRC someone else had this issue and the reason was that for whatever reason the route didn't use the rot layout from phoenix and phoenix is adding the csrf info in the layout. But I'm not sure exactly what caused it.

MaziyarMK:
found the problem: (this was done by igniter install into existing project)
```
url = url(~p"/auth/user/magic_link/?token=#{params[:token]}")
```
changed to

```
url = url(~p"/magic_link/#{params[:token]}")
```
in `accounts/user/senders/send_magic_link_email.ex`

---

Maybe this endpoint needs to be disabled if `magic_sign_in_route` is used? idk.
```
GET   /auth/user/magic_link          XYZ.Accounts.User.magic_link :accept
```

PR #981, opened 2025-05-02
improvement: pass along the tenant claim(if exists) to the long-lived token

adamtharani:
This change passes along the tenant from the sign_in token claims(if it exists) to the long-lived token claims that is returned to the AuthController.


zachdaniel:
@jimsynz this seems like a good approach to me, but needs your say too 🙇

jimsynz (approved):
nice work

PR #984, opened 2025-05-05
feat: Add API key authentication strategy

### LINKED ISSUE #855: Add an API token strategy.
API tokens are essentially "if you have this value then we can authenticate you", so it should be relatively simple to add an api token strategy. No need for a JWT to be involved, so it would have to make the jwt in the token resource optional, or be its own resource. I think perhaps `ApiToken` being its own resource wouldn't be too bad.

zachdaniel:
closes #855


jimsynz (approved):
ship it!

lib/ash_authentication/strategies/api_key/actions.ex
```diff
@@ -0,0 +1,75 @@
+defmodule AshAuthentication.Strategy.ApiKey.Actions do
+  @moduledoc """
+  Actions for the magic link strategy.
```
maennchen, line 3: Should mention API keys and not magic links

lib/ash_authentication/strategies/api_key/actions.ex
```diff
@@ -0,0 +1,75 @@
...
+
+  Provides the code interface for working with user resources for providing
+  magic links.
+  """
+
+  alias Ash.{Query, Resource}
+  alias AshAuthentication.{Errors, Info, Strategy.ApiKey}
+
+  @doc """
+  Attempt to sign a user in via magic link.
```
maennchen, line 13: Magic Links

lib/ash_authentication/strategies/api_key/plug.ex
```diff
@@ -0,0 +1,185 @@
...
+  This plug validates API keys from either a query parameter or HTTP header.
+  """
+
+  @behaviour Plug
+
+  alias Ash.PlugHelpers
+  alias AshAuthentication.{Info, Strategy}
+  alias Plug.Conn
+
+  @type source_type :: :header | :query_param
```
maennchen, line 14:
```suggestion
  @type source_type :: :header | :query_param | :header_or_query_param
```

lib/ash_authentication/strategies/api_key/actions.ex
```diff
@@ -0,0 +1,75 @@
...
+
+      {:ok, _users} ->
+        {:error,
+         Errors.AuthenticationFailed.exception(
+           strategy: strategy,
+           caused_by: %{
+             module: __MODULE__,
+             strategy: strategy,
+             action: :sign_in,
+             message: "Query returned too many users"
```
maennchen, line 51: I would not differentiate between no users or too many users.

If two users use the same insecure token, we don't want them to know that deleting theirs will let them access the other one.

Also there should probably be a unique index by default preventing this in the first place.

zachdaniel: This error isn't returned to the end user, it is only used internally (and yeah there is an identity added to protect about this).

But I could also hide this information even from the developer if that is more secure?

documentation/tutorials/api-keys.md
```diff
@@ -0,0 +1,122 @@
...
+    data_layer: AshPostgres.DataLayer,
+    authorizers: [Ash.Policy.Authorizer]
+
+  actions do
+    defaults [:read, :destroy]
+
+    create :create do
+      primary? true
+      accept [:api_key, :user_id, :expires_at]
+    end
```
maennchen, line 44: I would probably deter users from being able to chose their own tokens. Could we remove `api_key` from `accept` by default and provide a change that chooses a random token with an ok amount of entropy?

lib/ash_authentication/strategies/api_key/actions.ex
maennchen: If it's only internal, it's probably fine.

lib/ash_authentication/strategies/api_key/generate_api_key.ex
```diff
@@ -0,0 +1,54 @@
...
+
+  @base62 Enum.with_index(Enum.concat([?a..?z, ?A..?Z, ?0..?9]))
+
+  use Ash.Resource.Change
+
+  def change(changeset, opts, _) do
+    prefix = Keyword.fetch!(opts, :prefix)
+    rand_bytes_count = Keyword.get(opts, :rand_bytes) || 32
+    token = :crypto.strong_rand_bytes(rand_bytes_count)
+    id_generator = opts[:id_generator] || (&Ash.UUID.generate/0)
```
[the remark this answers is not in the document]
zachdaniel: My thought was that they might want to use something like UUIDv7, but I'll just take this out for now and force uuid everywhere. We can iterate.

documentation/tutorials/api-keys.md
```diff
@@ -0,0 +1,122 @@
+# API Keys
+
+## A note on API Keys
+
+API keys are a bit different from the other strategies in AshAuthentication, in that it is your job to create and manage them.
+
+This means that to leverage them in the standard way, you will need to build a UI for managing API keys. You can create API keys in iex in development or if you don't have a need for users to create their own API keys.
+
+It is up to you to determine what constitutes a valid API key, and who is allowed to create/modify them. All that we require is that an API key is a string attribute on the related api key resource. Be sure that the API key relationship used with this strategy only includes valid API keys.
```
maennchen, line 9: Needs to be adapted to the new code.

documentation/tutorials/api-keys.md
```diff
@@ -0,0 +1,123 @@
...
+    data_layer: AshPostgres.DataLayer,
+    authorizers: [Ash.Policy.Authorizer]
+
+  actions do
+    defaults [:read, :destroy]
+
+    create :create do
+      primary? true
+      accept [:user_id, :expires_at]
+      change {AshAuthentication.Strategy.ApiKey.GenerateApiKey, prefix: :my_app, hash: :api_key_hash}
```
maennchen, line 44: We don’t allow underscores in the prefix.

```suggestion
      change {AshAuthentication.Strategy.ApiKey.GenerateApiKey, prefix: :myapp, hash: :api_key_hash}
```

lib/ash_authentication/strategies/api_key/sign_in_preparation.ex
```diff
@@ -0,0 +1,114 @@
...
+      _api_keys ->
+        {:error,
+         AssumptionFailed.exception(
+           message: "Multiple API tokens found for actor matching: #{inspect(api_key)}"
+         )}
+    end
+  end
+
+  defp decode_api_key(api_key) do
+    with [_parse_prefix, middle, crc32] <- String.split(api_key, "_"),
```
maennchen, line 102: Avoid parsing very far and creating large lists if user provides manipulate tokens in the request.

```suggestion
    with [_parse_prefix, middle, crc32] <- String.split(api_key, "_", parts: 3),
```

documentation/tutorials/api-keys.md
```diff
@@ -0,0 +1,123 @@
...
+The API key is generated using a random byte string and a prefix. The prefix is used to generate a key that is compliant with secret scanning. You can use this to set up an endpoint that will automatically revoke leaked tokens, which is an extremely powerful and useful security feature. We only store a hash of the api key. The plaintext api key is only available in `api_key.__metadata__.plaintext_api_key` immediately after creation.
+
+See [the guide on Github](https://docs.github.com/en/code-security/secret-scanning/secret-scanning-partnership-program/secret-scanning-partner-program) for more information.
+
+Api key expiration/validity is otherwise up to you. The configured `api_key_relationship` should include those rules in the filter.
+For example:
+
+```elixir
+has_many :valid_api_keys, MyApp.Accounts.ApiKey do
+  filter expr(is_expired == false)
```
maennchen, line 15: The calculation is called `valid` below.

```suggestion
  filter expr(valid == false)
```

documentation/tutorials/api-keys.md
```diff
@@ -0,0 +1,123 @@
...
+      allow_nil? false
+    end
+  end
+
+  calculations do
+    calculate :valid, :boolean, expr(expires_at > now())
+  end
+
+  identities do
+    identity :unique_api_key, [:api_key]
```
maennchen, line 73:
```suggestion
    identity :unique_api_key, [:api_key_hash]
```

lib/ash_authentication/strategies/api_key/plug.ex
```diff
@@ -0,0 +1,185 @@
...
+    - Otherwise, the response will be:
+      - Status: 403 Forbidden
+      - Content-Type: text/plain (default)
+      - Body: `Forbidden`
+  """
+  def on_error(conn, _error) do
+    if Plug.Conn.get_req_header(conn, "accept") |> Enum.any?(&String.contains?(&1, "json")) do
+      conn
+      |> Plug.Conn.put_resp_content_type("application/json")
+      |> Plug.Conn.send_resp(403, ~s({"error":"Forbidden"}))
```
maennchen, line 38: An invalid token usually is 401...

```suggestion
      |> Plug.Conn.send_resp(401, ~s({"error":"Unauthorized"}))
```

lib/ash_authentication/strategies/api_key/plug.ex
```diff
@@ -0,0 +1,185 @@
...
+    end
+  end
+
+  @doc """
+  Initialize the plug with options.
+
+  ## Options
+
+  * `:resource` - The resource to authenticate against.
+  * `:source` - Where to get the API key from. Can be `:header`, `:query_param` or `:header_or_query_param`. Default: `:header`
```
maennchen, line 53: Maybe do a callout that query_params have a tendency to be stored in logs and ask the user to choose wisely...

lib/ash_authentication/strategies/api_key/sign_in_preparation.ex
```diff
@@ -0,0 +1,114 @@
...
+      )
+      |> Query.after_action(fn
+        _query, [record] ->
+          verify_hash_and_set_api_key(record, query, api_key_id, random_bytes, strategy, api_key)
+
+        _query, [] ->
+          Plug.Crypto.secure_compare(
+            :crypto.hash(:sha256, random_bytes <> api_key_id),
+            Ecto.UUID.bingenerate() <> :crypto.strong_rand_bytes(32)
+          )
```
maennchen, line 36: The timing is off for this branch since there's no second `Ash.load`.
Is there a way to do both queries in one to avoid the difference?

zachdaniel: We don't actually have a great way to load associations in a single query in Ash, we've avoided it as (except for this one scenario now) its never really been anything other than an optimization. There are things that can be done using calculations, but that is fairly difficult to make generic in an extension like this... 🤔 I don't want to make a fake query to the database with a random UUID either though 😓.

🤔

lib/ash_authentication/strategies/api_key/plug.ex
```diff
@@ -0,0 +1,186 @@
...
+  end
+
+  @doc """
+  Initialize the plug with options.
+
+  ## Options
+
+  * `:resource` - The resource to authenticate against.
+  * `:source` - Where to get the API key from. Can be `:header`, `:query_param` or `:header_or_query_param`. Default: `:header`.
+    Keep in mind tha tquery params are often stored in logs etc, so we *highly* recommend using `:header`.
```
maennchen, line 54:
```suggestion
    Keep in mind that query params are often stored in logs etc, so we *highly* recommend using `:header`.
```

maennchen (approved):

Issue #346, opened 2023-07-06
Document Auth0 log out process


guglielmobartelloni:
When a user tries to logut from a phoenix application using auth0 strategy for ash, the session will be cleared but auth0 is not notified about the logout. 

A workaround that I found is to redirect the user to `https://[YOUR_APP_ON_AUTH0]/v2/logout?returnTo=http%3A%2F%2F[YOUR_APP_HOME_PAGE]` endpoint when loggin out. This is a sample code from my authcontroller
```elixir
  def sign_out(conn, _params) do

    conn
    |> clear_session()
    |> redirect(external: "https://[auth0_endpoint]/v2/logout?returnTo=#{AppWeb.Endpoint.url()}")
  end

```
It would be helpful if the ash_authentication feature could automatically initiate a call to the auth0 endpoint. Additionally, it would be beneficial to include this workaround in the documentation, as new users of ash might not be aware of this functionality, just like myself.

jimsynz:
I think that we should add this to the documentation and maybe expose a function to generate the URL for you from the Auth0 strategy.

bg-numarqe:
Also, worth ensuring the `client_id` is the URL as well (necessary for when there are multiple apps).  As documented at https://community.auth0.com/t/new-to-auth0-allowed-url-logout-not-working-in-my-next-js-app/123824.

```
redirect(external: "https://[auth0_endpoint]/v2/logout?client_id=[auth0_client_id]&returnTo=#{AppWeb.Endpoint.url()}")
```

PR #985, opened 2025-05-07
doc: logout process for auth0

guglielmobartelloni:
Add documentation for the logout process when using Auth0 #346


jimsynz (approved):
thank you

PR #986, opened 2025-05-07
improvement(Confirmation): Provide a default accept phase form.

jimsynz:
When `require_interaction?` is turned on the `GET` for that conirmation route no longer works.  This change adds an accept phase similar to the one provided by the magic link strategy.


zachdaniel:
Doesn't this create a get request with a side effect then? Isn't that what we're trying to avoid by requiring interaction?

jimsynz:
No, it creates a get request which renders the "confirm" button and submits via a POST.

zachdaniel:
Oh right.

zachdaniel (approved):

PR #993, opened 2025-05-15
improvement(Strategy.Custom): The `strategy_module` field is no longer required.

### LINKED ISSUE #992: Clarify needing to define `strategy_module` and `provider` as part of custom strategy struct
The [custom strategy docs](https://hexdocs.pm/ash_authentication/custom-strategy.html#dsl-setup) only mention that a `resource` key is required in the struct, but `strategy_module` and `provider` appear to be required as well.

`strategy_module` is mentioned here - https://github.com/team-alembic/ash_authentication/blob/main/lib/ash_authentication/strategies/custom.ex#L22-L23 but is this accurate? This also doesn't mention `resource` or `provider`.

Can we confirm exactly which fields are needed, and make sure they *are* in fact all needed, and update the docs to reflect this?

jimsynz:
It's only needed when you recycle another strategy's DSL entity as your own. This is done a lot by all the OAuth2 and OIDC strategies which use the same OAuth2 struct behind the scenes.

Closes #992.

### DIFF
--- a/documentation/dsls/DSL-AshAuthentication.Strategy.Slack.md
+++ b/documentation/dsls/DSL-AshAuthentication.Strategy.Slack.md
@@ -8,7 +8,7 @@ Strategy for authenticating using [Slack](https://slack.com)
 This strategy builds on-top of `AshAuthentication.Strategy.Oidc` and
 [`assent`](https://hex.pm/packages/assent).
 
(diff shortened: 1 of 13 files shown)


sevenseacat (approved):
awesome, that clears it all up! 👍

PR #998, opened 2025-05-19
feat(Password): Support password hashing with Argon2.

jimsynz:
We still default to Bcrypt for now. This will change in the next major release.


zachdaniel:
IIRC argon2 is non trivially slower than bcrypt unless you configure it to be faster, maybe we should have a benchmark? Something unrelated that we'll want eventually is a version that doesn't need a NIF, and we will likely want the installer to detect windows and use that one by default. phx.gen.auth does something similar (@maennchen told me about it recently).

maennchen:
@zachdaniel I’m not quite sure what you’re referring to…

In any case: Argon2 being slow is a feature and not an issue 😊 It’s slow on purpose to make brute-forcing the hash impossible.

zachdaniel:
IIRC you mentioned phx gen auth detects windows and uses a native erlang hashing algorithm, but maybe it was someone else 😄

zachdaniel:
> In any case: Argon2 being slow is a feature and not an issue 😊 It’s slow on purpose to make brute-forcing the hash impossible.

I was just thinking to make sure that it or the implementation wasn't problematically slow.

jimsynz:
I think (and maybe @maennchen can confirm?) that argon2 is the recommended algorithm these days. Bcrypt was long in the tooth when I chose it, but I knew the trade offs of it and trusted it. Honestly, the way we wrote it we can support a whole bunch of hashing algorithms out of the box, but choosing one by default puts our thumb on the scales of how safe our user community is.

jimsynz:
I think (and maybe @maennchen can confirm?) that argon2 is the recommended algorithm these days. Bcrypt was long in the tooth when I chose it, but I knew the trade offs of it and trusted it. Honestly, the way we wrote it we can support a whole bunch of hashing algorithms out of the box, but choosing one by default puts our thumb on the scales of how safe our user community is.

maennchen:
> phx gen auth detects windows and uses a native erlang hashing algorithm

I wasn’t aware, but that sounds very reasonable.

> I was just thinking to make sure that it or the implementation wasn't problematically slow.

That makes sense, the defaults should be sane I think. But if you actually have hundreds or more of logins a second, it can be problematic. I would leave it up to the user to choose how to deal with that in this instance though.

For tests, I would lower the settings to make them fast.

> I think […] that argon2 is the recommended algorithm these days. 

Don’t trust me for this, but for example here you can see the OWASP recommendations:

https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html

maennchen:
Also, Bryan just kicked off a discussion for native Argon2 support in the EEF security WG channel:
https://the-eef.slack.com/archives/CTP7P1E9X/p1747590805273849

jimsynz:
Yeah so based on OWASP Argon2id is the way to go. I think we should also add support for PBKDF2 for folks needing FIPS-140. 

I don't quite understand the explanation of peppering so I'll read some more but that seems like something we'd want to support too.

PR #1000, opened 2025-05-21
Return the new_conn instead of the old conn

adonig:
Fixes an issue with the testing example code, which assigns a `new_conn` but then wrongly returns the old `conn`:

```elixir
    new_conn =
      conn
      |> Phoenix.ConnTest.init_test_session(%{})
      |> AshAuthentication.Plug.Helpers.store_in_session(user)

   %{context | conn: conn}  # <-- Error is here, should be new_conn
```


zachdaniel:

PR #1007, opened 2025-05-26
Info.strategy_for_action/2 -> Info.find_strategy/3

Torkan:
Use `Info.find_strategy` for looking up the strategy instead of `Info.strategy_for_action` directly.

The motivation for this change is that `ash_events` does some "clever" wrapping/renaming of actions when tracking events on resources.

When it is used on a resource that is also set up with `AshAuthentication`, the names of the various sign_in-actions get changed, and `Info.strategy_for_action` blows up since there is no strategy defined for the action after it has been renamed to something else than for example `sign_in_with_magic_link`.

The solution is to add a `change set_context(%{strategy_name}, <strategy>)` to the sign-in actions when using `ash_events`, but this will only work if the strategy is determined through `Info.find_strategy` 😅 

For context:
https://github.com/ash-project/ash_events/pull/20


zachdaniel (approved):

zachdaniel:

PR #1010, opened 2025-05-29
improvement: add helpful error messages for invalid secret return values

### LINKED ISSUE #437: Helpful Error from `AshAuthentication.Secret` `secret_for`
Trying to set up Auth0 using [this guide](https://ash-hq.org/docs/guides/ash_authentication/latest/tutorials/auth0-quickstart) I got to the login error page, but no errors were raised.

Here is the reason from the `AuthController`:

```elixir
reason #=> %AshAuthentication.Errors.MissingSecret{
  resource: Red.Accounts.User,
  changeset: nil,
  query: nil,
  error_context: [],
  vars: [],
  path: [:authentication, :strategies, :auth0, :client_id],
  stacktrace: #Stacktrace<>,
  class: :forbidden
}
```

Eventually I tried this:

```elixir
  def secret_for([:authentication, :strategies, :auth0, :redirect_uri], Red.Accounts.User, _) do
    get_config(:redirect_uri)
    |> dbg()
  end
```

Which showed me this:

`get_config(:redirect_uri) #=> :error`

Only here did I realize I made the mistake of saving the wrong key in my config file `redirect_url`

What I think should happen, is that if my `secret_for` function does not return the correct type, a helpful error is raised, letting me know what I returned.

jimsynz:
Resolves #437 by implementing better error handling in the secret_for function. When secret functions return unexpected values (not strings, lists, maps, or booleans), developers now get clear error messages indicating the actual returned value instead of generic "MissingSecret" errors.

- Add InvalidSecret error with descriptive messages
- Enhance AshAuthentication.Secret.secret_for/5 to validate return values
- Simplify JWT config error handling (now redundant)
- Add comprehensive tests for various invalid secret scenarios

Example error message:
"Secret for `authentication.strategies.oauth2.redirect_uri` on the `Example.User` resource returned an invalid value. Expected a string, list, map, or boolean but got nil."

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/lib/ash_authentication/secret.ex
+++ b/lib/ash_authentication/secret.ex
@@ -62,6 +62,7 @@ defmodule AshAuthentication.Secret do
   """
 
   alias Ash.Resource
+  alias AshAuthentication.Errors.InvalidSecret
 
   @doc deprecated: "Use AshAuthentication.Secret.secret_for/4 instead"
   @callback secret_for(secret_name :: [atom], Resource.t(), keyword) :: {:ok, String.t()} | :error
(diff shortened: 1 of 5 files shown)


barnabasJ (approved):

lib/ash_authentication/secret.ex
```diff
@@ -124,10 +125,38 @@ defmodule AshAuthentication.Secret do
...
+    result =
+      if module.__secret_for_arity__() == 4 do
+        module.secret_for(secret_name, resource, opts, context)
+      else
+        module.secret_for(secret_name, resource, opts)
+      end
+
+    case result do
+      {:ok, secret}
+      when is_binary(secret) or is_list(secret) or is_map(secret) or is_boolean(secret) ->
```
zachdaniel, line 137: I don't really understand why a secret should be any one of these particular types? I think we should just raise when the secret doesn't return `ok/error` tuple and let secrets return whatever they want.

PR #1026, opened 2025-06-16
improvement: add utility to revoke all session tokens

zachdaniel:
improvement: update `revoke_bearer_tokens` to take `opts`

### DIFF
--- a/lib/ash_authentication/plug/helpers.ex
+++ b/lib/ash_authentication/plug/helpers.ex
@@ -17,8 +17,8 @@ defmodule AshAuthentication.Plug.Helpers do
     if Info.authentication_tokens_require_token_presence_for_authentication?(user.__struct__) do
       Conn.put_session(conn, "#{subject_name}_token", user.__metadata__.token)
     else
-      subject = AshAuthentication.user_to_subject(user)
-      Conn.put_session(conn, subject_name, subject)
+      {:ok, %{"sub" => subject, "jti" => jti}} = Jwt.peek(user.__metadata__.token)
+      Conn.put_session(conn, subject_name, jti <> ":" <> subject)
     end
   end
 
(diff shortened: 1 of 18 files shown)


lib/ash_authentication/plug/helpers.ex
```diff
@@ -267,23 +267,69 @@ defmodule AshAuthentication.Plug.Helpers do
...
+            conn
+
+          _ ->
+            conn
+        end
+
+      {_resource, _options, false}, conn ->
+        # we can't revoke tokens for resources that don't store them
+        # because the only thing in the session is the subject
+        conn
```
zachdaniel, line 329: This is not the default setup. With that said, @jimsynz I think that perhaps we could add an option to store the token in the session and not just the subject, even when `require_token_presence_for_authentication?` is not enabled, then this use case would be enabled.

zachdaniel: Going to revisit this before merge. We're currently storing just the subject which is not realistically enough to revoke the token on sign out, which is something we have to be able to do regardless of the strategy at play here.

zachdaniel:
This PR will log out all users for applications that are not using `require_token_presence_for_authentication` because, in order to revoke tokens on log-out we have to store the jti in the subject.

lib/ash_authentication/token_resource/actions.ex
```diff
@@ -281,6 +281,38 @@ defmodule AshAuthentication.TokenResource.Actions do
     end
   end
 
+  @doc """
+  Revoke a token by JTI.
+
+  If you have the token, you should instaed use `revoke/2` instead.
```
barnabasJ, line 287:
```suggestion
  If you have the token, you should use `revoke/2` instead.
```

jimsynz (approved):
SECURITAY!

PR #1031, opened 2025-07-01
Send opts to validate_token

josepjaume:
In a multitenant environment, the token wasn't properly found as `opts` weren't being sent to `validate_token` and thus `tenant` wasn't being set. This fixes it.


zachdaniel:

PR #1033, opened 2025-07-01
fix(retrieve_from_session): incorrectly assuming tokens are stored.

jimsynz:
When `store_all_tokens?` is `false` but `require_token_presence_for_authentication?` is `true`, the plug helper was performing an extra token resource check that is already verified by the Jwt.verify/3 function and not required.

This caused an infinite redirect loop back to the sign up page on some apps.


zachdaniel:
Should that configuration combination  even be possible?

jimsynz:
Of course. All `require_token_presence_for_authentication?` does is enable the usage of tokens in the session, rather than just assuming that because the session says it's a user we should believe it

jimsynz:
There won't be. But there'll be revocation records if they're revoked.

mbuhot (approved):

zachdaniel:
I'm still pretty confused how you'd ever login if you require tokens storage and also don't store tokens...

jimsynz:
> I'm still pretty confused how you'd ever login if you require tokens storage and also don't store tokens...

Okay yeah, going back and reading the docs for `require_token_presence_for_authentication?` seems like it's a synonym for `store_all_tokens?` so maybe we should update this PR to set `store_all_tokens?` to whatever the other thing is set to and deprecate it.

zachdaniel:
🤔 doesn't it also require that the token be in the token resource too? 
- `require_token_presence_for_authentication? true` - require that the token exists in the database. Also changes from storing subject in the session to storing the whole token
- `store_all_tokens? true` - store all generated tokens in the token resource

It would seem like the first just requires the second to be `true` to make sense.  if you can use the first without the second then I'm confused.

zachdaniel:
🤔interesting. Okay. Why would it cause an infinite redirect to check again? 

Separately, seeing a change to arguably one of the most critical checks we have (happy to accept that it is wrong) in addition to changes in the assertions around that behavior.

jimsynz:
I'm going to close this PR and come back with another take.

zachdaniel:
I'm sure you're right in the change you want to make I'm just being like 10000% careful in my review to make sure I understand.

PR #1037, opened 2025-07-05
DRAFT: improve: API key auth for multi-tenancy resources

serpent213:
Introduce optional `multitenancy_relationship` to the API key strategy to allow the authenticated resource to be bound to a tenant.

This is a  follow-up to https://elixirforum.com/t/ashauthentication-with-multi-tenancy-and-api-keys-strategy/

With this patch I can do:
```elixir
defmodule User do
  multitenancy do
    strategy :context
  end

  authentication do
    strategies do
      api_key :api_key do
        api_key_relationship :valid_api_keys
        api_key_hash_attribute :api_key_hash
        multitenancy_relationship :organisation
      end
    end
  end
  # [...]
end
```

Might this be a valid approach?

### DIFF
--- a/lib/ash_authentication/strategies/api_key.ex
+++ b/lib/ash_authentication/strategies/api_key.ex
@@ -27,7 +27,8 @@ defmodule AshAuthentication.Strategy.ApiKey do
             resource: nil,
             sign_in_action_name: nil,
             api_key_hash_attribute: nil,
-            api_key_relationship: nil
+            api_key_relationship: nil,
+            multitenancy_relationship: nil
 
   alias AshAuthentication.Strategy.{ApiKey, ApiKey.Transformer, ApiKey.Verifier}
 
@@ -38,7 +39,8 @@ defmodule AshAuthentication.Strategy.ApiKey do
           resource: Ash.Resource.t(),
           sign_in_action_name: atom(),
           api_key_hash_attribute: atom(),
-          api_key_relationship: atom()
+          api_key_relationship: atom(),
+          multitenancy_relationship: atom()
         }
 
   @doc false
--- a/lib/ash_authentication/strategies/api_key/dsl.ex
+++ b/lib/ash_authentication/strategies/api_key/dsl.ex
@@ -33,6 +33,10 @@ defmodule AshAuthentication.Strategy.ApiKey.Dsl do
           type: :atom,
           doc:
             "The name to use for the sign in action. Defaults to `sign_in_with_<strategy_name>`"
+        ],
+        multitenancy_relationship: [
+          type: :atom,
+          doc: "The relationship from the API key to the issuing tenant."
         ]
       ]
     }
--- a/lib/ash_authentication/strategies/api_key/sign_in_preparation.ex
+++ b/lib/ash_authentication/strategies/api_key/sign_in_preparation.ex
@@ -24,6 +24,7 @@ defmodule AshAuthentication.Strategy.ApiKey.SignInPreparation do
         api_key_relationship.destination
         |> Ash.Query.do_filter(api_key_relationship.filter)
         |> Ash.Query.filter(id == ^api_key_id)
+        |> maybe_load_tenant(strategy.multitenancy_relationship)
         |> Query.set_context(%{private: %{ash_authentication?: true}})
         |> Ash.read_one()
         |> case do
@@ -77,6 +78,7 @@ defmodule AshAuthentication.Strategy.ApiKey.SignInPreparation do
         api_key_relationship.source_attribute =>
           Map.get(api_key, api_key_relationship.destination_attribute)
       })
+      |> maybe_set_tenant(api_key, strategy.multitenancy_relationship)
       |> Ash.Query.after_action(fn
         _query, [user] ->
           {:ok,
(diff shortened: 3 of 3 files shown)


serpent213:
Of course I want to recycle the tenant in my plug to set the scope – `__metadata__.tenant` should provide that. Apparently the tenant is not carried forward when being set in the `before_action`...?

I added a second commit to set it explicitly in the `after_action`. Tbh, I don't fully understand why that is necessary. I also tried replacing the `set_metadata` with two calls to `put_metadata`, but the result was the same...

serpent213:
Tests are LLM generated. Look ok to me.

serpent213:
> For setting tenant in the metadata here, I think you'll need to do it in an `after_action` hook and map over the results setting the tenant.

Not sure what you mean...

zachdaniel:
You mentioned this:

> Apparently the tenant is not carried forward when being set in the before_action...?

To solve that you'd need to add an after action hook that loops over the results and uses `Ash.Resource.set_metadata` to set the tenant into the records that are returned.

serpent213:
Computer says **no**.  😅

Sounds reasonable to me. Maybe it's just fine to handle this edge case (set_tenant in before_action) as is?

serpent213:
**The Solution: Use the Correct Query Struct**

To fix this, we need to replace all instances of query with new_query in the do_run's with block and its do block after the data has been fetched. The new_query variable holds the correct state of the query after all hooks and transformations inside do_read.

Here is the corrected do_run function. The changes are commented.

```elixir
defp do_run(query, action, opts) do
  # ... (initial part of the function is the same) ...
  # ...

  initial_query = query
  # ... (middle part of the function is the same) ...
  # ...
  else
    try do
      data_result =
        # ... (this part is the same) ...

      {data_result, query_ran} =
        case data_result do
          {:ok, _result, _count, _calculations_at_runtime, _calculations_in_query, query} =
              data_result ->
            {data_result, query}

          {{:error, _} = data_result, query} ->
            {data_result, query}

          data_result ->
            {data_result, query}
        end

      query = Ash.Query.set_context(query, %{shared: query_ran.context[:shared]})

      with {:ok, data, count, calculations_at_runtime, calculations_in_query, new_query} <-
             data_result,
           # FIX: Use `new_query` which has the state from after the before_action hook
           data = add_tenant(data, new_query),
           {:ok, data} <-
             load_through_attributes(
               data,
               %{query_ran | calculations: Map.new(calculations_in_query, &{&1.name, &1})},
               query.domain,
               opts[:actor],
               opts[:tracer],
               opts[:authorize?]
             ),
           # FIX: Use `new_query` for loading relationships
           {:ok, data} <-
             load_relationships(data, new_query, opts),
           {:ok, data} <-
             Ash.Actions.Read.Calculations.run(
               data,
               new_query, # FIX: Use `new_query` for calculations
               calculations_at_runtime,
               calculations_in_query
             ),
           {:ok, data} <-
             load_through_attributes(
               data,
               %{
                 new_query # FIX: Use `new_query` here as well
                 | calculations: Map.new(calculations_at_runtime, &{&1.name, &1}),
                   load_through: Map.delete(query.load_through || %{}, :attribute)
               },
               query.domain,
               opts[:actor],
               opts[:tracer],
               opts[:authorize?],
               false
             ) do
        data
        # FIX: Use `new_query` for all subsequent processing
        |> Helpers.restrict_field_access(new_query)
        |> add_tenant(new_query)
        |> attach_fields(opts[:initial_data], initial_query, new_query, missing_pkeys?)
        |> cleanup_field_auth(new_query)
        |> add_page(
          new_query.action, # FIX: Use `new_query`
          count,
          new_query.sort, # FIX: Use `new_query`
          new_query, # FIX: Use `new_query`
          new_query,
          opts
        )
        |> add_query(query, opts) # `add_query` uses the original query for different purposes, this can stay
      else
        # ... (error handling part is the same) ...
      end
    after
      # ... (after block is the same) ...
    end
  end
end
```

By making these changes, you ensure that the final processing steps, including add_tenant, operate on the query struct that has been fully modified by all the preceding steps, including your before_action hook. This will correctly add the tenant key to the __metadata__ of your returned records.

serpent213:
After a thorough review of `update.ex`, the good news is that it does not suffer from the same variable scoping issue as read.ex. The way it handles the changeset lifecycle is different and correctly uses the state of the changeset after hooks are applied.

Let's break down why.

**update.ex vs. read.ex: A Structural Comparison**

The core difference lies in how the main logic block is structured and where the final record processing happens.

* In read.ex (The Problematic One):
   1. The do_run function calls do_read.
   2. do_read runs the before_action hooks and returns a new_query struct with the updated state.
   3. Execution returns to do_run, which then proceeds to process the results, but it mistakenly uses the original query variable from its own scope, ignoring the new_query that contained the tenant information.
* In update.ex (The Correct One):
   1. The do_run function calls commit.
   2. The commit function uses Ash.Changeset.with_hooks to wrap the entire data layer operation.
   3. Ash.Changeset.with_hooks is responsible for running all the hooks (before_transaction, before_action, etc.) and passing the modified changeset into the main logic block.
   4. The crucial step of attaching the tenant to the resulting record happens inside this main logic block, using the correctly modified changeset.

serpent213:
That would indeed work!

serpent213:
Ready to merge, as far as I can tell.

Sorry for the LLM slop – happy I could contribute to the new guidelines!  😉

PR #1038, opened 2025-07-07
improve: allow regex header prefix for ApiKey.Plug

skanderm:
Allows passing a regex header_prefix in the ApiKey.Plug

### DIFF
--- a/lib/ash_authentication/strategies/api_key/plug.ex
+++ b/lib/ash_authentication/strategies/api_key/plug.ex
@@ -198,7 +198,7 @@ defmodule AshAuthentication.Strategy.ApiKey.Plug do
   defp validate_header_prefix!(prefix) when is_binary(prefix), do: :ok
   defp validate_header_prefix!(%Regex{source: "^" <> _rest}), do: :ok
 
-  defp validate_header_prefix!(%Regex{source: regex_string}) do
-    raise "Invalid header_prefix regex. Regexes must begin with `^`, got: #{regex_string}"
+  defp validate_header_prefix!(%Regex{} = regex) do
+    raise "Invalid header_prefix regex. Regexes must begin with `^`, got: #{inspect(regex)}"
   end
 end


lib/ash_authentication/strategies/api_key/plug.ex
```diff
@@ -138,8 +138,10 @@ defmodule AshAuthentication.Strategy.ApiKey.Plug do
   defp get_api_key(conn, %{source: :header, header_prefix: prefix}) do
     case Conn.get_req_header(conn, "authorization") do
       [header_value | _] ->
-        if String.starts_with?(header_value, prefix) do
-          {:ok, String.replace_prefix(header_value, prefix, "")}
+        prefix = Regex.compile!("^" <> prefix, "i")
```
zachdaniel, line 141: 🤔 this is actually an interesting question, I'm not sure if this should be case insensitive by default or not, but perhaps instead we allow the prefix to be a regex? That would allow this behavior without making it the default.

skanderm: Fair enough! Looks like it's supposed to be case sensitive: https://datatracker.ietf.org/doc/html/rfc6750#section-1.1. I mostly added this because I was wondering why my requests were erroring as unauthorized.

I like the idea of passing in a regex though. Happy to change this.

test/ash_authentication/strategies/api_key_test.exs
```diff
@@ -71,6 +71,21 @@ defmodule AshAuthentication.Strategy.ApiKeyTest do
       assert conn.assigns.current_user.id == user.id
     end
 
+    test "succeeeds when API key is present in header with regex prefix", %{
+      plaintext_api_key: plaintext_api_key,
+      user: user
+    } do
+      opts = ApiKeyPlug.init(resource: Example.User, source: :header, header_prefix: ~r/bearer/i)
```
zachdaniel, line 78: This is technically a bug I believe, as you'd want ^bearer. Perhaps we should validate that the regex starts with `^`

skanderm: Oops, good catch! 😬

I've updated with a validation and a test. Let me know if you'd like any other changes!

lib/ash_authentication/strategies/api_key/plug.ex
```diff
@@ -66,10 +66,12 @@ defmodule AshAuthentication.Strategy.ApiKey.Plug do
     required? = Keyword.get(opts, :required?, true)
     source = Keyword.get(opts, :source, :header)
     param_name = Keyword.get(opts, :param_name, "api_key")
-    header_prefix = Keyword.get(opts, :header_prefix, "Bearer ")
     subject_name = AshAuthentication.Info.authentication_subject_name!(resource)
     assign = Keyword.get(opts, :assign, :"current_#{subject_name}")
 
+    header_prefix = Keyword.get(opts, :header_prefix, "Bearer ")
+    :ok = validate_header_prefix(header_prefix)
```
zachdaniel, line 73: Last suggestion is not to match on the return here but just call it, and add a `!` to the function.

jimsynz, line 73: I agree with Zach's suggestion here - since the function raises on invalid input, it would be more idiomatic to name it `validate_header_prefix!` and just call it without matching on the return value.

test/ash_authentication/strategies/api_key_test.exs
```diff
@@ -71,6 +71,33 @@ defmodule AshAuthentication.Strategy.ApiKeyTest do
       assert conn.assigns.current_user.id == user.id
     end
 
+    test "succeeeds when API key is present in header with regex prefix", %{
```
jimsynz, line 74: Typo: "succeeeds" should be "succeeds"

lib/ash_authentication/strategies/api_key/plug.ex
```diff
@@ -183,4 +190,14 @@ defmodule AshAuthentication.Strategy.ApiKey.Plug do
       context: context
     )
   end
+
+  defp validate_header_prefix(prefix) when is_binary(prefix), do: :ok
+
+  defp validate_header_prefix(%Regex{} = prefix) do
```
jimsynz, line 196: The regex validation approach using string inspection is a bit fragile, but I can't think of a better way to do it either 🤷‍♂️ This works fine for the use case.

jimsynz (commented):
A few minor suggestions to address:

lib/ash_authentication/strategies/api_key/plug.ex
skanderm: Absolutely, great suggestion

PR #1049, opened 2025-07-29
improvement: add claims to JWT by Context

marcnnn:
This change allows adding claims to magic links by passing them in context

Documentation is missing because I think the Ash Team will have opinions about this change.

After finding the correct solution that works for everyone, I will add the Documentation.

### DIFF
--- a/lib/ash_authentication/strategies/magic_link.ex
+++ b/lib/ash_authentication/strategies/magic_link.ex
@@ -191,12 +191,15 @@ defmodule AshAuthentication.Strategy.MagicLink do
   """
   def request_token_for_identity(strategy, identity, opts \\ [], context \\ %{})
       when is_struct(strategy, __MODULE__) do
+        context_map = Map.from_struct(context)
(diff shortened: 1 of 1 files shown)


lib/ash_authentication/strategies/magic_link.ex
```diff
@@ -191,12 +191,15 @@ defmodule AshAuthentication.Strategy.MagicLink do
   """
   def request_token_for_identity(strategy, identity, opts \\ [], context \\ %{})
       when is_struct(strategy, __MODULE__) do
+        context_map = Map.from_struct(context)
+        source_context = Map.get(context_map, :source_context, %{})
+        context_claims = Map.get(source_context, :claims, %{})
```
zachdaniel, line 196: Hm...I think we'd have to stew on this one. Likely if we do this the name here would need to be much more specific though, like `:magic_link_token_claims` for example.

@jimsynz thoughts?

PR #1062, opened 2025-09-04
Create action validation behaviour, make module configurable in DSL.

Torkan:
This PR defines a behaviour `AshAuthentication.Validations.ActionValidators`, matching the implementation of `AshAuthentication.Validations.Action`,  and adds the `action_validators` DSL-field so that another validation module can be configured if needed, like this:

```
authentication do
  action_validators AshEvents.AshAuthenticationActionValidators
end
```

Furthermore, all previous direct usage of `AshAuthentication.Validations.Action` has been replaced by fetching and using the configured module.

The main reason for this PR is that other extensions (in this particular case, `ash_events`) might need to make modifications to an action's changes, and since `ash_authentication`'s verifiers that use `AshAuthentication.Validations.Action` expects to find the action changes in a specific way, the verifiers will fail even though the functionality of the added change remains intact.

Other extensions that manipulate an action in such a manner, can then define their own validators module to use with `ash_authentication`, that users can utilize if they are using both extensions in the same project.

This PR is 100% backwards-compatible and requires no effort from existing users of `ash_authentication`.

### DIFF
--- a/lib/ash_authentication/add_ons/confirmation/transformer.ex
+++ b/lib/ash_authentication/add_ons/confirmation/transformer.ex
@@ -5,13 +5,11 @@ defmodule AshAuthentication.AddOn.Confirmation.Transformer do
   Ensures that there is only ever one present and that it is correctly
   configured.
   """
-
   alias Ash.{Resource, Type}
   alias AshAuthentication.{AddOn.Confirmation, GenerateTokenChange}
   alias Spark.{Dsl.Transformer, Error.DslError}
   import AshAuthentication.Utils
   import AshAuthentication.Validations
-  import AshAuthentication.Validations.Action
   import AshAuthentication.Validations.Attribute
   import AshAuthentication.Strategy.Custom.Helpers
 
@@ -172,11 +170,16 @@ defmodule AshAuthentication.AddOn.Confirmation.Transformer do
   end
 
   defp validate_confirm_action(dsl_state, strategy) do
-    with {:ok, action} <- validate_action_exists(dsl_state, strategy.confirm_action_name),
-         :ok <- validate_action_has_change(action, Confirmation.ConfirmChange),
-         :ok <- validate_action_argument_option(action, :confirm, :allow_nil?, [false]),
-         :ok <- validate_action_argument_option(action, :confirm, :type, [Type.String]),
-         :ok <- validate_action_has_change(action, GenerateTokenChange) do
+    action_validator = AshAuthentication.Info.authentication_action_validators!(dsl_state)
+
+    with {:ok, action} <-
+           action_validator.validate_action_exists(dsl_state, strategy.confirm_action_name),
+         :ok <- action_validator.validate_action_has_change(action, Confirmation.ConfirmChange),
+         :ok <-
+           action_validator.validate_action_argument_option(action, :confirm, :allow_nil?, [false]),
+         :ok <-
+           action_validator.validate_action_argument_option(action, :confirm, :type, [Type.String]),
+         :ok <- action_validator.validate_action_has_change(action, GenerateTokenChange) do
       accept_fields = MapSet.new(action.accept || [])
 
       strategy.monitor_fields
--- a/lib/ash_authentication/add_ons/log_out_everywhere/transformer.ex
+++ b/lib/ash_authentication/add_ons/log_out_everywhere/transformer.ex
@@ -12,7 +12,6 @@ defmodule AshAuthentication.AddOn.LogOutEverywhere.Transformer do
   alias Spark.{Dsl.Transformer, Error.DslError}
   import AshAuthentication.Utils
   import AshAuthentication.Validations
-  import AshAuthentication.Validations.Action
   import AshAuthentication.Strategy.Custom.Helpers
 
   @doc false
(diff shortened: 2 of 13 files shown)


barnabasJ:
> > > Not really sure why the commit message check fails, when I run it locally it doesn't complain at all 🤔 🤷
> > 
> > 
> > I think it might check the pull request title as well
> 
> Aha, thanks! I'll try updating and re-running then 😅

Ok, that wasn't it. But also not too important. 

What changes are you trying to make with AshEvents exactly? Overwriting the whole validator feels a bit heavy-handed.
I want to understand why the validator fails in your case, and if we can just make the validator a bit more robust.

Torkan:
In order for AshEvents to work properly, it needs to "wrap" all changes in its ReplayChangeWrapper, so there is no way AshAuthentication can validate that the change is present, unless we make it aware of how AshEvents alter the action (not a good approach 😅)

It's really just `validate_action_has_change` that is causing the issue, but it's trivial to implement the functions you need in a module that has the `AshAuthentication.Validations.ActionValidators` behaviour for, then one can just use `defdelegate` in order to just use the functions from `AshAuthentication.Validations.Action` for all the others.

zachdaniel:
That check is broken unfortunately. I'll leave this one to @barnabasJ or @jimsynz to review, but generally I'm in support 😄

barnabasJ:
> Possibly yes. I hadn't considered doing it at the spark level. But that also makes it hard to grow together over time. It means that new validations won't be picked up etc.

I'm not 100% sure I get what you mean, how would that differ from doing it this way, in regards to new validations not getting picked up?

Torkan:
Yes, the main risk factor is that the verifier is rewritten to for example not use the configured validator module anymore, and do this check inline, which would break AshEvents. Also AshEvents would need to keep its action validator module in sync with the behaviour that AshAuthentication defines.

This would be the case for any other Ash extensions that utilizes a custom action validator module as well.

For AshAuthentication in isolation though, there is no particular downside (at least as far as I can see 😅), besides the effort of rewriting the internal code to fetch the configured module, which this PR already takes care of.

I'm totally fine with the risk mentioned above on behalf of AshEvents, and it'll probably be extremely few (if any) other extensions that would need to leverage this functionality, so it shouldn't be a huge burden to be compliant to keep other extensions compliant with the ActionValidators behaviour overall.

If an even simpler/more elegant solution comes to mind however, then I'm all for it! 
I just can't see any other angles to this myself 😅

barnabasJ:
Like I said, I would prefer a more generic solution in Spark, like a config, where you can specify a replacement for a validator, for example. 

To me, it's not about the downsides for AshAuthentication, more about the pattern/precedence this sets.

That way, you could just do what you want for every extension without having to necessarily involve/convince the maintainers that they need to expose their validators. 

Having said that, I will let @jimsynz decide 😅

zachdaniel:
> That way, you could just do what you want for every extension without having to necessarily involve/convince the maintainers that they need to expose their validators.

 I don't believe that this is a good thing at all. The point is that wholesale replacing another extensions validators is like using a private API. This is AshAuthentication providing a *public* API to determine how its validations ought to be applied. No other extension I know of currently does this kind of valuation.

What we could do instead as a middle ground is create something like `Ash.Resource.Validator` that lives in Ash core that we then migrate `AshAuthentication` to use, and then tell people who want to have their validations overwrite able to support taking an option to specify a validator module.

zachdaniel:
@barnabasJ curious to hear your thoughts. It seems like a heavy thing to add to core for this one particular use case. @Torkan IIRC we also had some kind of concept of saying what the "canonical action" for a given action was, for extensions that rewrite actions. That would theoretically require extensions to know to ask for that and validate that though. So... 🤷 I'm not sure.

barnabasJ:
> I don't believe that this is a good thing at all. The point is that wholesale replacing another extensions validators is like using a private API. This is AshAuthentication providing a public API to determine how its validations ought to be applied. No other extension I know of currently does this kind of validation.
> 
> What we could do instead as a middle ground is create something like Ash.Resource.Validator that lives in Ash core that we then migrate AshAuthentication to use, and then tell people who want to have their validations overwrite able to support taking an option to specify a validator module.

The heart of the problem for me is that the validator is something private either way. And it's not that AshAuthentication wants its Validator to be overwritable, it gains nothing by doing that (looking at it in isolation, otherwise, of course, you can say it gains compatability with AshEvents). In my mind, as an extension author, the only thing I'm gaining by opting into having my validator overwritable, are probably more Github issues because of things not working as expected because of validator differences. 

I can also see other Extensions validating actions in the future, some of the already existing extensions that add changes to actions could theoretically already do it, but may not have found it necessary. 

Having it in Spark and making it something that you need to explicitly configure feels more like a framework (spark) mechanism (escape hatch) to me, it also moves the responsabilty more towards the extension doing the overwriting. 

From a growing together perspective, I think you mostly have the same problems regardless. As long as the releases are not done in sync in both cases, you could have situations where you upgrade only one of the extensions, and it breaks your build. The behaviour specified in the first approach would make it a tad more stable, I give you that, and it would make it possible to specify the overwrite on a Resource basis, which can be a pro.

In the end, I think this would be fine as a one-of, but if this is something more extensions need to do in the future, I just don't feel like extensions exposing their validators is a wholeistic solution to this problem.

PR #1064, opened 2025-09-07
improve: ensure_get_by_action uses get_by

RicoTrevisan:
I was going to fat-finger a pr to get `ensure_get_by_action` uses `get_by` instead of `get?` & `filter`. But the tests fail because the output is actually `get_by(:email)` instead of the expected `get_by :email` . 

First, is it intentional that `ensure_get_by_action` is set up the way it is? Or is there something wrong with my local setup that is generating that out?
CleanShot 2025-09-07 at 11 49 01@2x


zachdaniel:
Yeah, for "reasons" the tests don't use the formatter. Don't worry about it, just go with it 😆

RicoTrevisan:
I've realized I could remove the args altogether. I realized that I need to improve the title and some documentation of the PR.

zachdaniel (approved):

PR #1068, opened 2025-09-14
fix: Add only existing sections into Ash.Resource

aifrak:
Currently when running Igniter, it adds `:authentication` and `:tokens` under `Ash.Resource` in `config.exs`. The main issue is that `:tokens` do not seem to exist or have been moved under  `:authentication`. Also the documentation mentions another section called `:user_identity` but is not added by Igniter.

This PR removes the non-existing section `:tokens` and adds the missing section `:user_identity` in the config under `Ash.Resource`.

I just tried out `ash_authentication`, so I am not so familiar with it. Let me know if something is not correct.

### DIFF
--- a/lib/mix/tasks/ash_authentication.install.ex
+++ b/lib/mix/tasks/ash_authentication.install.ex
@@ -91,7 +91,7 @@ if Code.ensure_loaded?(Igniter) do
       |> Igniter.Project.Config.configure("test.exs", :bcrypt_elixir, [:log_rounds], 1)
       |> Spark.Igniter.prepend_to_section_order(
         :"Ash.Resource",
-        [:authentication, :tokens]
+        [:authentication, :user_identity]
       )
       |> Igniter.compose_task(
         "ash.gen.domain",


jimsynz:
You're half right. The `AshAuthentication.TokenResource` extension _does_ define the `:token` DSL, but we did miss the `:user_identity` DSL from the igniter. I'm not really sure what the right thing here is, because I think that both the `token` and `user_identity` DSL sections are likely to be one of only a very few DSL sections in their respective files.

zachdaniel:
The order we provide is just a suggestion, just putting them in there is all that we really need to do 😄

aifrak:
> The `AshAuthentication.TokenResource` extension does define the `:token` DSL, [...]

Thanks for the feedback. Sorry, I missed that this one is only in `AshAuthentication.TokenResource`. I have added `:token` (singular).

> I'm not really sure what the right thing here is, because I think that both the token and user_identity DSL sections are likely to be one of only a very few DSL sections in their respective files.

I also hesitated for the same reason. But I think it would be nice to add them in order to show the new sections this extension is adding without diving into the documentation. Also, as they are only 3, I think it is fine to add them all.

Let me know if it is better to remove them and I will adjust my PR to have only the common section, which is only `:authentication`.

You can review my PR again.

jimsynz (approved):
details

PR #1074, opened 2025-10-03
feat: Audit Logging add-on.

jimsynz:
This PR brings a wonderful new add-on to AshAuthentication - audit logging!

I know that doesn't sound very exciting, but I think it unlocks heaps of new and interesting use cases and features for us.

For example:

```
calculate :currently_locked_out?, :boolean, expr(count(MyApp.Account.AuditLog, email == parent(:email) and action == :sign_in_with_password and audit_log == :failed_sign_in_tracker and logged_at >= ago(15, :minute)) > 3)
```

or

```
calculate :mark_for_deletion?, :boolean, expr(not exists(MyApp.Account.AuditLog, email == parent(:email) and logged_at >= ago(90, :day))
```

I considered whether or not to use `ash_events` to implement the log storage, however decided;
1. I didn't want to bring an additional dependency into `ash_authentication` and as `AshAuthentication.AuditLogResource` is an extension that just defines attributes and actions it's up to the user if they want to back it with `ash_events`
2. This isn't just a log of successful changes, but all actions whether they succeed or fail.

## Release blockers

- [x] blocked by https://github.com/ash-project/ash/pull/2352/files is released because the password strategy generator generates a generic `:request_password_reset_token` action without a return type.
- [x] Ash dependency bumped after :arrow_up: is released.


lib/ash_authentication/add_ons/audit_log/auditor.ex
```diff
@@ -0,0 +1,170 @@
...
+      AshAuthentication.Info.strategy_for_action!(input.resource, input.action.name)
+
+    status =
+      case result do
+        ok when elem(ok, 0) == :ok -> :success
+        error when elem(error, 0) == :error -> :failure
+        :error -> :failure
+        {ok, _} when elem(ok, 0) == :ok -> :success
+        {error, _} when elem(error, 0) == :error -> :failure
+        _ -> :unknown
```
barnabasJ, line 95: I think this could be written more ideomatic and I'm not sure if these lines

```elixir
        {ok, _} when elem(ok, 0) == :ok -> :success
        {error, _} when elem(error, 0) == :error -> :failure
```

are necessary? I can't see those in the type definition https://github.com/ash-project/ash/blob/b107cb0d4b59e9e69212ad6c5e129051c5620688/lib/ash/subject.ex#L43-L43

lib/ash_authentication/audit_log_resource/batcher.ex
```diff
@@ -0,0 +1,191 @@
...
+        Map.put(configs, resource, config)
+    end)
+  end
+
+  defp write_batch(resource, config) do
+    bulk_result =
+      config.queue
+      |> Stream.map(& &1.attributes)
+      |> Ash.bulk_create(resource, config.action,
+        authorize?: false,
```
barnabasJ, line 172: should we really do this? shouldn't this be handled by the policy that checks the context?

barnabasJ: But no policies are added by the igniter/transformer at the moment, maybe we should add it there and remvoe the authorize?: false here.

jimsynz: Yeah, this definitely should use the ash_authentication context key.

lib/ash_authentication/add_ons/audit_log/auditor.ex
jimsynz: You were right. This is definitely a bug. https://github.com/ash-project/ash/pull/2354

lib/ash_authentication/add_ons/audit_log/auditor.ex
```diff
@@ -0,0 +1,189 @@
...
+          &Auditor.after_transaction(&1, &2, opts[:strategy], context)
+        )
+      else
+        changeset
+      end
+    end
+
+    @doc false
+    @impl true
+    def atomic(changeset, opts, context), do: {:ok, change(changeset, opts, context)}
```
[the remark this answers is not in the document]
jimsynz: @zachdaniel can you comment? Is this correct?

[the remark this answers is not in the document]
zachdaniel: Nope, this doesn't mean that nothing is logged if it's atomic. It means that the atomic and non-atomic behavior is the same. We allow after_transaction hooks on atomic actions as well (since we have a slightly more loose definition of atomicity then true atomicity).

maennchen (approved):
Awesome :)

PR #1086, opened 2025-11-21
feat: Add TOTP authentication strategy.

### LINKED ISSUE #503: Support TOTP as a second factor.

jimsynz:
Implements TOTP (Time-based One-Time Password) authentication strategy for AshAuthentication.

Fixes ALE-11
Closes #503

## Summary

- Add TOTP strategy DSL and transformer
- Add core changes: GenerateSecretChange, SignInPreparation, VerifyAction
- Add Actions module for setup, sign_in, and verify
- Add Plug module for HTTP handling
- Add QR code URL calculation via otpauth:// URIs
- Add two-step setup confirmation (generate → verify → save)
- Add replay attack prevention via last_totp_at timestamp
- Add comprehensive test suite (461 tests)
- Add documentation (tutorial + DSL docs)

## Key Features

- Both standalone TOTP and 2FA modes (after password)
- All three brute force protection strategies (rate limit, audit log, custom prep)
- Two-step setup with confirmation (generate → verify → save)
- QR code generation for authenticator apps
- Replay attack prevention via timestamp tracking

## Test plan

- [x] All 461 tests passing
- [x] `mix check` passes
- [x] Manual testing with authenticator app

## Links
- Linear Project: https://linear.app/teamalembic/project/multi-factor-authentication-aaa1d659b49e
- Linear Issue: https://linear.app/teamalembic/issue/ALE-11
- Phoenix PR: https://github.com/team-alembic/ash_authentication_phoenix/pull/706


lib/ash_authentication/strategies/magic_link/transformer.ex
barnabasJ: This change was probably not intentional?

lib/ash_authentication/strategies/totp/audit_log_change.ex
```diff
@@ -0,0 +1,118 @@
...
+    logged_at_attr = AuditLogResource.Info.audit_log_attributes_logged_at!(audit_log_resource)
+
+    query =
+      audit_log_resource
+      |> Ash.Query.new()
+      |> Ash.Query.filter(^ref(subject_attr) == ^subject)
+      |> Ash.Query.filter(^ref(strategy_attr) == :totp)
+      |> Ash.Query.filter(^ref(status_attr) == :failure)
+      |> Ash.Query.filter(^ref(logged_at_attr) >= ^cutoff)
+      |> Ash.Query.lock("FOR UPDATE")
```
barnabasJ: Do audit logs have update actions? I guess we don't enforce them not having one? 

Just wondering what the lock exactly is for, the count wouldn't change even if the records were updated.

lib/ash_authentication/strategies/totp/confirm_setup_change.ex
```diff
@@ -0,0 +1,199 @@
...
+      {:ok, []} ->
+        {:error, invalid_token_error(strategy, "Setup token not found or expired")}
+
+      {:ok, _tokens} ->
+        {:error, invalid_token_error(strategy, "Ambiguous setup token")}
+
+      :error ->
+        {:error, invalid_token_error(strategy, "Invalid setup token")}
+
+      {:error, reason} ->
```
barnabasJ: Above you had another match in a similar case where you checked if it was an exception and in that case passed it through otherwise wrap it in an exception. Maybe you want to the same here?

lib/ash_authentication/strategies/totp/dsl.ex
```diff
@@ -0,0 +1,170 @@
...
+             ]},
+          doc: "How you are mitigating brute-force token checks.",
+          required: true
+        ],
+        identity_field: [
+          type: :atom,
+          doc:
+            "The name of the attribute which uniquely identifies the user, usually something like `username` or `email_address`.",
+          default: :username
+        ],
```
barnabasJ: this is the only field that doesn't have `required` set at all, should we also add it here with the correct value to make it consistent?

lib/ash_authentication/strategies/totp/audit_log_change.ex
jimsynz: The lock prevents a race condition where multiple concurrent TOTP attempts could slip past the brute force limit. Without it, several parallel requests could all count N failures (where N is just under the limit), all pass the check, and then all log new failures - exceeding the intended limit.

The lock serialises the count queries so each request sees an accurate failure count. I've added an inline comment explaining this and consolidated the code into `AuditLogHelpers.count_failures/3`.

lib/ash_authentication/strategies/magic_link/transformer.ex
jimsynz: Fixed - changed the period to a comma.

test/ash_authentication/strategies/totp_test.exs
```diff
@@ -0,0 +1,134 @@
...
+      {:ok, signed_in_user} =
+        Strategy.action(
+          strategy,
+          :sign_in,
+          %{strategy.identity_field => user_with_secret.email, "code" => code},
+          []
+        )
+
+      # Reload user from database to check last_totp_at
+      {:ok, reloaded_user} = Ash.get(Example.UserWithTotp, signed_in_user.id, authorize?: false)
```
barnabasJ: Just curious, why is the reload needed?

lib/ash_authentication/strategies/magic_link/transformer.ex
jimsynz: Fixed in 2bb8315 - it was an unintentional comma-to-period change.

lib/ash_authentication/strategies/totp/actions.ex
```diff
@@ -0,0 +1,226 @@
...
+               field: :user,
+               type: :argument
+             )
+           ]
+         )}
+
+      user ->
+        options =
+          options
+          |> Keyword.put_new_lazy(:domain, fn -> Info.domain!(strategy.resource) end)
```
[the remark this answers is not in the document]
jimsynz: Same - addressed in 2bb8315.

lib/ash_authentication/strategies/totp/audit_log_preparation.ex
```diff
@@ -0,0 +1,160 @@
...
+         {:ok, audit_log} <- get_audit_log(input.resource, strategy),
+         user when not is_nil(user) <- ActionInput.get_argument(input, :user) do
+      check_rate_limit_for_input(input, user, strategy, audit_log, opts, context)
+    else
+      _ -> input
+    end
+  end
+
+  defp check_rate_limit_after_action(_query, results, strategy, audit_log, opts, context) do
+    case results do
```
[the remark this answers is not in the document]
jimsynz: This was partially addressed in 2bb8315 - the shared code was extracted to `AuditLogHelpers`. There's still some structural similarity between Change and Preparation, but they have different interfaces (`change/3` vs `prepare/3`) and handle results differently. Happy to consolidate further if you think it's worthwhile, but the current split keeps each module focused on its specific callback pattern.

lib/ash_authentication/strategies/totp/totp_url_calculation.ex
```diff
@@ -0,0 +1,99 @@
...
+
+    Enum.map(records, fn record ->
+      build_totp_url(record, strategy)
+    end)
+  end
+
+  defp build_totp_url(record, strategy) do
+    secret = Map.get(record, strategy.secret_field)
+
+    if is_nil(secret) or secret == "" do
```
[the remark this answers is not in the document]
jimsynz: Acknowledged - thanks for the follow-up.

jimsynz:
Added `read_secret_from` DSL option (3d42b36) to support AshCloak integration. This allows specifying a calculation to read the decrypted TOTP secret from, while `secret_field` is used for writing the encrypted value.

Issue #892, opened 2025-01-23
Upgrade of Assent in 4.4.2 breaks Google Auth


allenwyma:
This commit: abe1a759c5524e58b985e93a24d03c315b3553fe has broken `ash_authenticate` 4.2.2 due to Assent being automatically upgraded to 0.3.0 which breaks the current integration for Google.

zachdaniel:
Sorry about this, fixed in `4.2.3`

PR #1090, opened 2025-11-23
chore: update all deps

joshprice:
Removed too specific version rules. Should remove all warnings when installing as a dep.


joshprice:
@shahryarjb `jose` is not a direct dep, so not sure what is happening there. Looks like it comes in via `assent` and `joken`.

joshprice:
@shahryarjb looks like ash_auth_phoenix is locked to 1.11.11

https://github.com/team-alembic/ash_authentication_phoenix/blob/main/mix.lock#L42

joshprice:
More likely to have broken in 1.11.10 -> 1.11.11 in that case.

Is there some way to reproduce your failing tests as a test here?

Will be tricky to track down otherwise.

shahryarjb:
Hi dear @joshprice i sent a pr for it
https://github.com/team-alembic/ash_authentication/pull/1092
Thank you

allenwyma:
hi @joshprice updating assent to 0.3 will break google logins.

Please see: [#892](https://github.com/team-alembic/ash_authentication/issues/892)

PR #1092, opened 2025-11-23
fix: support :null atom from JOSE 1.11.11+ in JWT tenant validation

shahryarjb:
`JOSE` 1.11.11+ changed how nil values are handled when using Erlang/OTP 27+'s built-in json module.
JSON null is now decoded as the :null atom instead of Elixir's nil, causing tenant claim validation to fail for master/global tokens (where tenant is nil). 
This results in 403 Forbidden errors for users with nil tenant (for example site_id nil). 

This fix `normalizes :null` to nil before comparison to support both old and new JOSE versions.

> Handles both `nil` and `:null` atom values to support JOSE 1.11.11+ which decodes JSON `null` as the `:null` atom when using Erlang's built-in json module.

Diff: https://diff.hex.pm/diff/jose/1.11.10..1.11.11

> All tests passed in my project and `ash_authentication`, and i tested it in 2 version, the current and elixir 1.19 otp 28

Thank you in advance

### DIFF
--- a/lib/ash_authentication/jwt.ex
+++ b/lib/ash_authentication/jwt.ex
@@ -6,7 +6,7 @@ defmodule AshAuthentication.Jwt do
   @default_algorithm "HS256"
   @default_lifetime_days 7
   @supported_algorithms Joken.Signer.algorithms()
-  import AshAuthentication.Utils, only: [to_sentence: 2]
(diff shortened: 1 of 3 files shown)


jimsynz:
Thanks for the contribution!

PR #1098, opened 2025-11-28
fix!: Support Assent v0.3.0 OIDC-based Google strategy

### LINKED ISSUE #1097: Upgrade of assent breaks current Google auth login
https://github.com/team-alembic/ash_authentication/pull/1090#issuecomment-3587886494

Image

joshprice:
Users upgrading using Google auth strategy should be aware of these field name changes in the `user_info` map returned by Google:
  - `google_hd` → `hd` (hosted domain)
  - `email_verified` is now a `boolean` instead of a `string`

If user code accesses `user_info["google_hd"]`, it should be updated to `user_info["hd"]`."
  
Fixes #1097


maennchen (approved):
@joshprice I don't know the internals of the assent oidc strategy, but this implementation looks good to me.

zachdaniel:
This would be a breaking change then wouldn't it?

joshprice:
@zachdaniel yes it's a breaking change, but also it's a fix to the assent upgrade that is broken in main (the release it was in was retired)

PR #1105, opened 2025-12-06
fix(audit_log): invalid magic links log as success

### LINKED ISSUE #1104: AshAuth audit log magic link failures are logged as successes
A failed magic link log in attempt create a log with `status == :success`, instead of `status == :failure`

rgraff:
When a magic link token is invalid, it does not return a user but also does not return an error. Therefore the request is logged as a success by the audit logger. Fixes #1104


jimsynz (approved):
celebration!

PR #1112, opened 2025-12-27
Fix error caused by after_action ordering

capoccias:
I've seen the following error in production:
```elixir
* ** (FunctionClauseError) no function clause matching in AshAuthentication.Strategy.Password.SignInWithTokenPreparation.revoke_sign_in_token/4

lib/ash_authentication/strategies/password/sign_in_with_token_preparation.ex in AshAuthentication.Strategy.Password.SignInWithTokenPreparation.revoke_sign_in_token/4 at line 74

arg0	[Filtered]
arg1	[]
arg2	[Filtered]
arg3	[Filtered]
```

This happens when the query returns no users which is a case handled by `&verify_result/4` but not `&revoke_sign_in_token/4`.

In order to reproduce the bug I had to set Ash config value, `read_action_after_action_hooks_in_order?` to true as is suggested by https://github.com/ash-project/ash/blob/main/documentation/topics/development/backwards-compatibility-config.md#read_action_after_action_hooks_in_order. I'm not sure if it's best to keep this enabled or not but the fix I implemented handles both cases.

### DIFF
--- a/config/test.exs
+++ b/config/test.exs
@@ -17,8 +17,9 @@ config :ash_authentication, Example.Repo,
 config :bcrypt_elixir, :log_rounds, 4
 config :argon2_elixir, t_cost: 1, m_cost: 8
 
-config :ash, :disable_async?, true
-config :ash, :read_action_after_action_hooks_in_order?, true
+config :ash,
+  disable_async?: true,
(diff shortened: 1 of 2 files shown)


jimsynz (approved):
order matters

PR #1131, opened 2026-02-04
improvement: add `remember_me` to password sign-in strategy by default

zachdaniel:
remember_me was configured, but wasn't being shown in forms on new applications because the argument/change wasn't there on the action.


jimsynz (approved):
lest we forget

PR #1134, opened 2026-02-08
improvement: add `grace_period` option to TOTP strategy

jimsynz:
## Summary

- Adds a `grace_period` DSL option to the TOTP strategy that accepts codes from previous time periods, implementing the [pattern recommended by NimbleTOTP docs](https://hexdocs.pm/nimble_totp/NimbleTOTP.html#valid?/3-grace-period)
- When set to `n`, validates against the current period plus `n` previous periods. Defaults to `nil` (no grace period, no behaviour change)
- Centralises all `NimbleTOTP.valid?/3` calls through a new `Helpers.valid_totp?/4` function

## Test plan

- [x] `mix compile --warnings-as-errors` — clean
- [x] All 22 TOTP tests pass
- [x] `mix check --no-retry` — all checks pass


adamtharani (approved):
Tested this out, and it worked as expected

PR #1136, opened 2026-02-19
feat: add microsoft strategy

MikaelFangel:
I use `ash_authentication` for my Phoenix projects, and I really like it. I couldn't see a Microsoft strategy, so I thought I would add one and I hope this is a welcome improvement. I've made the tutorial and DSL so it's quite similar to other strategies.

There were a few catches:

- It was needed to remove the form_post response mode to avoid triggering the CSRF in the default browser pipeline from Phoenix.
- I couldn't find a clean way to use tenant_id directly because the discovery endpoint path is either `/common/v2.0` or `/{tenant_id}/v2.0` depending on your app registration, so I opted for allowing the user to override with the full `base_url` instead.

### DIFF
--- a/lib/ash_authentication/strategies/microsoft/dsl.ex
+++ b/lib/ash_authentication/strategies/microsoft/dsl.ex
@@ -39,15 +39,6 @@ defmodule AshAuthentication.Strategy.Microsoft.Dsl do
       auto_set_fields: [icon: :microsoft, assent_strategy: AzureADMultitenant]
     })
     |> Custom.set_defaults(AzureADMultitenant.default_config([]))
-    |> Map.update!(
-      :schema,
-      fn schema ->
-        # Override response mode from form_post to avoid CSFR
-        Keyword.update!(schema, :authorization_params, fn config ->
-          Keyword.put(config, :default, scope: "email profile")
-        end)
-      end
-    )
   end
 
   defp strategy_override_docs(strategy) do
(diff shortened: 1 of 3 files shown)


jimsynz:
Thanks for putting this together\! A Microsoft strategy is something we've been wanting — really keen to get this merged. I'll do a code review shortly.

lib/ash_authentication/strategies/microsoft/dsl.ex
```diff
@@ -0,0 +1,70 @@
...
+      - The [Microsoft Tutorial](/documentation/tutorials/microsoft.md)
+      - The [OIDC documentation](`AshAuthentication.Strategy.Oidc`)
+
+      #### Strategy defaults:
+
+      #{strategy_override_docs(AzureADMultitenant)}
+      """,
+      auto_set_fields: [icon: :microsoft, assent_strategy: AzureADMultitenant]
+    })
+    |> Custom.set_defaults(AzureADMultitenant.default_config([]))
```
jimsynz: The `strategy_override_docs/1` function generates from `AzureADMultitenant.default_config([])` which includes `response_mode: "form_post"`. But the schema override on line 51 deliberately strips `form_post` from the default. So the generated cheat sheet advertises `response_mode: "form_post"` as a default when it has actually been removed.

Either the docs generation should reflect the post-override defaults, or `strategy_override_docs` should be adjusted to exclude `authorization_params` (since it's overridden below).

PR #1141, opened 2026-03-15
Add OTP strategy

Torkan:
Adds a first-class OTP (one-time password) authentication strategy to AshAuthentication. Users receive a short code (e.g. `"XKPTMH"`) via email/SMS instead of a magic link URL, then submit it along with their identity to sign in.

How it works: The strategy derives a deterministic JWT ID from (strategy_name, user_subject, otp_code) via SHA-256. During request, it generates a random code, creates a JWT with that deterministic JTI, stores it in the existing TokenResource, and sends the short code to the user. During sign-in, it recomputes the JTI from the submitted code and looks it up — no new tables, actions, or schema changes to TokenResource needed.

Key features:
- Ambiguity-reduced default character set (excludes I/1, O/0, S/5, Z/2, etc.)
- Case-insensitive matching by default (configurable)
- Single-use codes by default (configurable)
- Optional registration support via registration_enabled? (upsert-based, like magic link)
- Pluggable generator module for custom code formats
- Auto-generated request and sign-in actions

DSL example:
```
strategies do
  otp do
    identity_field :email
    sender MyApp.OtpSender
  end
end
```


cospin:
Amazing, this is exactly what was missing from `ash_authentication` to support login with a mobile number the right way. Not sure if this is within the scope of this strategy or if it can be easily extended, but I agree with what @justinbkay mentions, it would be perfect to have a function to validate a `request` that isn’t part of a `sign_in` flow, usually on an already logged-in user, to use it with custom logic just as a validation.

For example, this new function would be more like a re-authentication for sensitive actions, 2FA, or even updating the identity, such as changing the mobile number to a new one while keeping the same account. If we use `sign_in` (to verify the new number before updating), a new account would be created, making it impossible to update the existing one since a previous account already exists with that identity.

Either way, thank you for the excellent work!

maennchen (changes requested):

jimsynz (commented):
## Feedback from the recovery code strategy (PR #1135)

The recently merged recovery code strategy solves many of the same problems the OTP strategy faces — both deal with short, limited-entropy codes that need brute-force protection, secure hashing, and single-use consumption. The main difference is that recovery codes are pre-computed and stored, while OTP codes are generated at request time. The infrastructure from #1135 is directly reusable here.

### 1. `brute_force_strategy` should be a required DSL option

The recovery code strategy requires users to declare their brute-force mitigation at the DSL level and validates it at compile time:

```elixir
brute_force_strategy: [
  type: {:or, [{:literal, :rate_limit}, {:tuple, ...}, {:tuple, ...}]},
  required: true
]
```

It supports `:rate_limit` (validates `AshRateLimiter` is configured), `{:audit_log, name}` (validates an audit log add-on is tracking the action), or `{:preparation, module}` (custom preparation). The verifier then enforces that the declared protection actually exists.

The OTP strategy currently has a warning in the tutorial docs, but no enforcement. OTP codes have *less* entropy than recovery codes (~26 bits for 6-char unambiguous uppercase vs ~62 bits for 12-char alphanumeric recovery codes), making them *more* vulnerable to brute force. Adopting the same required `brute_force_strategy` option would move this from "read the docs and hope" to "compile-time enforced".

### 2. Use `HashProvider` / `SHA256Provider` instead of hand-rolling crypto

The OTP strategy hand-rolls SHA-256:

```elixir
def compute_deterministic_jti(strategy, subject, normalized_otp) do
  data = "otp:#{strategy.name}:#{subject}:#{normalized_otp}"
  :sha256 |> :crypto.hash(data) |> Base.url_encode64(padding: false)
end
```

`AshAuthentication.SHA256Provider` (added by #1135) wraps exactly this and additionally provides:
- `Plug.Crypto.secure_compare` for constant-time comparison in `valid?/2`
- `minimum_entropy/0` for compile-time entropy validation
- `deterministic?/0` to enable atomic DB-level verification
- `simulate/0` for timing-attack resistance

Using the `HashProvider` behaviour would keep the hashing approach consistent across strategies and avoid reimplementing crypto.

### 3. Atomic token consumption instead of check-then-revoke

The recovery code strategy's `VerifyAction` shows the pattern for deterministic hash providers — a single atomic database operation with no TOCTOU race:

```elixir
strategy.recovery_code_resource
|> Ash.Query.filter(^ref(user_id_field) == ^user.id and ^ref(code_field) == ^hashed_input)
|> Ash.Query.limit(1)
|> Ash.bulk_destroy(:destroy, %{}, strategy: [:atomic, :atomic_batches, :stream], ...)
```

The OTP strategy's current sign-in path does three steps: `SELECT FOR UPDATE` → `jti_revoked?` → `revoke_jti`. The lock mitigates the race, but an atomic "find and consume in one operation" would eliminate it entirely and simplify the code. This also relates to @maennchen's concurrency feedback.

### 4. Entropy validation through `HashProvider.minimum_entropy/0`

The recovery code verifier validates entropy against the hash provider's declared minimum:

```elixir
entropy_bits = strategy.code_length * :math.log2(alphabet_size)
minimum = strategy.hash_provider.minimum_entropy()
```

The OTP verifier currently does its own check with a hardcoded `@min_otp_combinations 1_000_000` (~20 bits). If it used `SHA256Provider`, the minimum would be 60 bits — though the OTP use case is different (online attack surface with rate limiting vs offline brute-force of stored hashes), so the appropriate floor may differ. Either way, running entropy validation through the same `HashProvider` framework would keep the approach consistent. This also relates to @maennchen's entropy feedback.

### 5. Duplicated logic across sign-in paths

`sign_in_change.ex` and `sign_in_preparation.ex` both contain near-identical `get_otp_token_locked/3` and `maybe_consume_token` functions. Extracting these to a shared module would reduce the maintenance surface — especially if the consumption logic changes to adopt the atomic pattern above.

### Minor items

- `documentation/tutorials/otp.md` references `:unambiguous_digits` which doesn't exist in the DSL — should be `:unambiguous_alphanumeric` or removed.
- The tutorial's sender example logs OTP codes in plaintext — worth a note that this is for development only.
- Copyright header in `otp.md` says 2024.

Torkan:
Ok, adapted the strategy to use `brute_force_strategy` (required DSL option, same shape as `totp`/`recovery_code`). Also swapped the hand-rolled SHA-256 for `SHA256Provider.hash/1`, and deduplicated the two sign-in paths into `Otp.SignInHelpers`.

I didn't use `SHA256Provider.minimum_entropy/0` — its 60-bit floor would force 13–14 character OTP codes (or 19 digits), which defeats the point of OTP, so kept it as-is at 20.

Kept the `SELECT FOR UPDATE → check → revoke` pattern instead of atomic `bulk_destroy`. The `recovery_code` approach doesn't drop in cleanly because OTP reuses the shared `TokenResource`, which doesn't expose a plain destroy action (`expunge_expired` hard-codes its own filter).

A more proper fix could be adding a dedicated destroy action to `TokenResource`, that would also let `magic_link` move to atomic consumption as well? It appears to be more relaxed than this strategy currently on this front.

I also ensured `{:audit_log, _}` is rejected at compile time when `registration_enabled?: true`, since that needs  a user record to work.

Torkan:
@maennchen Should be ready for you to take a second look now, when you find the time 😄

maennchen:
@Torkan I'm a bit busy at the moment. But I assume based on my earlier feedback that my points have been addressed.

PR #1143, opened 2026-03-24
Split igniter generators into per-strategy and per-add-on tasks

jimsynz:
## Summary

- Extracts the monolithic `add_strategy` and `add_add_on` igniter tasks into composable sub-tasks (`ash_authentication.add_strategy.password`, `.magic_link`, `.api_key`, `ash_authentication.add_add_on.audit_log`, `.confirmation`)
- Parent tasks now validate and delegate via `Igniter.compose_task`, so existing usage (`mix ash_authentication.add_strategy password`) is unchanged
- Shared helpers added to `AshAuthentication.Igniter` as public API for use by third-party igniter tasks
- Confirmation add-on extracted from password strategy into its own task

## Test plan

- [x] All 26 existing igniter tests pass without modification
- [x] Clean compilation with `--warnings-as-errors`
- [ ] Verify `mix ash_authentication.add_strategy password` still works end-to-end in a fresh project
- [ ] Verify `mix ash_authentication.add_strategy.password` works directly
- [ ] Verify `mix ash_authentication.add_add_on audit_log` still works end-to-end


zachdaniel (commented):
Looks good to me! Just worth sanity checking that the ash_hq installer works after the change.

PR #1144, opened 2026-03-29
feat: add WebAuthn/Passkey authentication strategy

raul-gracia:
This PR adds a WebAuthn/Passkey authentication strategy to ash_authentication. Users can sign in with hardware security keys (YubiKey), platform authenticators (Touch ID, Windows Hello), and passkeys.

## What's in here

The strategy follows the same pattern as Password. A DSL entity defines configuration, a Transformer injects actions at compile time, a Verifier catches bad config early, an Actions module wraps the `wax_` library for FIDO2 ceremonies, and a Plug module handles the HTTP challenge/response flow.

New files:
- `lib/ash_authentication/strategies/webauthn.ex` - main module with struct and docs
- `lib/ash_authentication/strategies/webauthn/dsl.ex` - Spark DSL entity
- `lib/ash_authentication/strategies/webauthn/transformer.ex` - compile-time action injection
- `lib/ash_authentication/strategies/webauthn/verifier.ex` - config validation
- `lib/ash_authentication/strategies/webauthn/actions.ex` - registration, sign-in, credential management
- `lib/ash_authentication/strategies/webauthn/plug.ex` - HTTP handlers
- `lib/ash_authentication/strategies/webauthn/strategy.ex` - protocol implementation
- `lib/ash_authentication/strategies/webauthn/helpers.ex` - multi-tenant rp_id/origin resolution
- `lib/ash_authentication/strategies/webauthn/cose_key.ex` - custom Ash.Type for COSE keys
- `lib/ash_authentication/strategies/webauthn/sign_in_preparation.ex` - identity filter for read actions

## Configuration example

```elixir
authentication do
  tokens do
    enabled? true
    token_resource MyApp.Accounts.Token
    signing_secret fn _, _ -> {:ok, "my_secret"} end
  end

  strategies do
    webauthn :webauthn do
      credential_resource MyApp.Accounts.Credential
      rp_id "example.com"
      rp_name "My App"
      origin "https://example.com"
      identity_field :email
    end
  end
end
```

## Things worth noting

- `wax_` is an optional dependency. Only pulled in if you actually use WebAuthn.
- `origin` is separate from `rp_id`. Browsers include the port in origin, so dev on `https://localhost:4001` needs `origin "https://localhost:4001"` set explicitly. Default is `https://{rp_id}`, which works fine on standard ports.
- Challenges get stored in the Plug session as plain maps, not `Wax.Challenge` structs, because cookie stores can't serialize arbitrary structs.
- `register` creates a new user + credential. `add_credential` adds a key to an existing user.
- Credential management: list, rename, delete (with a guard preventing deletion of the last credential), and add.
- Multi-tenant support — rp_id, rp_name, and origin all accept MFA tuples.
- 43 new tests, 549 total (0 failures).

## Demo app

I built a demo Phoenix app to test the full browser flow: https://github.com/raul-gracia/webauthn_demo

Registration, sign-in, credential CRUD, and the add-key flow all work with real authenticators.

## Test plan

- [x] 43 automated tests (actions, plug, transformer, helpers, fixtures, management, multi-tenant)
- [x] Full test suite: 549 tests, 0 failures
- [x] Manual browser testing via demo app (registration, sign-in, key management)
- [x] Tested with Touch ID and Chrome virtual authenticator
- [x] Test with YubiKey hardware key
- [x] Test with Firefox and Safari


jimsynz:
Hi @raul-gracia. I'm excited to give this a review. Please be patient

jimsynz (commented):
Hey @raul-gracia — thank you so much for this contribution. WebAuthn/Passkey support has been on the wishlist for a long time, and you've done a tremendous amount of work to get us here. The strategy design follows the existing patterns really well (Transformer/Verifier/Actions/Plug), the test fixtures are clever (generating real EC key pairs for Wax verification without a browser), and the documentation in the main module is thorough. The multi-tenant support and credential management are nice touches too.

I've done a detailed audit of the code and wanted to share some findings. These aren't blockers on the direction — they're things that need addressing before we can ship it. Happy to pick some of these up myself if you'd prefer.

## Security items

### Nil challenge on missing/corrupt session (Medium)

`Plug.register/2` and `Plug.sign_in/2` call `reconstruct_challenge/3`, which returns `nil` when the session is empty or base64 decode fails. That `nil` flows through to `Actions.register/3` via `Keyword.fetch!(opts, :challenge)` (which succeeds — the key exists, the value is `nil`), and then into `Wax.register/3`, which will crash or return a confusing error.

An explicit nil guard before dispatching to the action would catch this cleanly:

```elixir
case reconstruct_challenge(conn, :attestation, strategy) do
  nil ->
    store_authentication_result(conn, {:error, AuthenticationFailed.exception(strategy: strategy)})
  challenge ->
    # proceed with registration
end
```

### Non-atomic registration — orphaned users (Medium)

`Actions.register/3` creates the user first (`create_user_from_registration`), then stores the credential separately (`store_credential_from_auth`). If the credential store fails (constraint violation, timeout, etc.), the user exists with no credential and no way to sign in. These should either be wrapped in a transaction or the credential creation should happen within the user's create action changeset.

### `Ash.update!` in best-effort code path (Low)

`update_sign_count/4` uses `Ash.update!` (the bang version) on the success path, but it's called from `best_effort_update_sign_count/4` which is supposed to be non-fatal. If the update raises, the sign-in fails despite successful Wax verification. Switching to `Ash.update` (non-bang) with a warning log would match the intended semantics.

### Bare rescue in `lookup_user_credentials` (Low)

The `rescue error ->` catches all exceptions. Worth narrowing to expected types so genuinely unexpected errors don't get silently swallowed.

## Functionality gaps

### No Igniter generator

The other strategies (Password, OAuth, MagicLink) have Igniter installers. WebAuthn has a lot of manual setup — adding the `wax_` dep, creating the credential resource with the right attributes/policies/relationship, adding the strategy DSL, running migrations, wiring up Phoenix components, and registering JS hooks. An Igniter generator would make adoption much more accessible.

### Credential resource not validated by the Transformer/Verifier

The Transformer validates the user resource's identity field, but doesn't check that the credential resource has the required attributes (`credential_id`, `public_key`, `sign_count`), a `belongs_to` relationship to the user, the necessary actions (`:create`, `:update`, `:read`, `:destroy`), or the `AshAuthenticationInteraction` policy bypass. Misconfigured credential resources will produce runtime errors instead of compile-time errors. Other strategies set a good precedent for catching these early.

### `add_credential` has no plug route

The action exists in `Actions` but `phases/1` and `routes/1` don't include it. It's documented as requiring a custom controller, but having a built-in phase for adding credentials to existing users would round out the feature — it's one of the most common WebAuthn flows after initial registration.

### `CoseKey.cast_stored/2` missing size limit

`cast_input/2` enforces the 256KB limit, but `cast_stored/2` (reading from the database) doesn't. Minor, but worth being consistent.

### No `AshAuthentication.Info` convenience functions

Other strategies expose helpers via the Info module. WebAuthn doesn't add strategy-specific convenience functions yet.

---

Again, massive thank you for getting this to such a solid state. The core ceremony handling is sound and the design fits the project well. If you'd like to tackle any of these items, brilliant — if not, I'm happy to pick them up and get it over the line. Let me know how you'd like to proceed.

raul-gracia:
Thanks @jimsynz for the thorough review, really appreciate the detailed audit and the offer to pick up the larger items.

I've just pushed a new commit (`c075916`) addressing the security items, and rebased the branch on top of `main` (v5.0.0-rc.3) to pick up the `usage_rules` bump that was failing CI.

## Addressed in this push

**Medium security:**

- **Nil challenge guard** — `Plug.register/2` and `Plug.sign_in/2` now explicitly match on the `nil` return from `reconstruct_challenge/3` and short-circuit via `store_authentication_result` with an `AuthenticationFailed` exception, instead of passing `nil` through to `Wax.register/3`.
- **Non-atomic registration** — wrapped the user creation and credential store in a compensating-cleanup pattern: if credential persistence fails after the user is created, the user is destroyed to prevent orphans. Cleanup failures are logged loudly (error level) but do not mask the original error. I went with compensating cleanup over a cross-resource transaction because it works regardless of the user/credential resources' data layers, but happy to switch to `Ash.transaction/2` or a changeset-embedded approach if you'd prefer.

**Low:**

- **`Ash.update!` → `Ash.update`** in `update_sign_count/4`, with a warning log on error. This was called from `best_effort_update_sign_count`, so a raise could kill a successful sign-in.
- **Narrowed the `rescue`** in `lookup_user_credentials/3` to `Ash.Error.Invalid`, `Ash.Error.Forbidden`, `Ash.Error.Framework`, and `Ash.Error.Unknown`, so genuine bugs (`KeyError`, `FunctionClauseError`, etc.) surface instead of being swallowed.

## Deferred (happy for you to take on)

- **Igniter generator** — this is substantial enough that I don't want to block shipping the core on it.
- **Credential resource validation in Transformer/Verifier** — agree this is important for DX.
- **`add_credential` plug phase/route** — currently in Actions only.
- **`CoseKey.cast_stored/2` size limit** — should mirror `cast_input/2`.
- **`AshAuthentication.Info` convenience functions.**

If it's easier I'm happy to open a follow-up PR for any of these, but totally understand if you'd rather take them on directly given the existing patterns in the codebase.

Locally all tests pass (584 tests, 0 failures) against the rebased branch, and credo/sobelow/format are clean. Let me know if there's anything else you'd like me to tweak.

jimsynz:
Superseded by #1159 — taken over to land the deferred review items (credential resource shape validation, `add_credential` plug route, `CoseKey.cast_stored/2` size limit) on top of your work. Massive thanks for the contribution — the strategy design and core ceremony handling carry through unchanged.

PR #1150, opened 2026-04-13
fix: remove dup migration name in password strategy and standardise

joshprice:
The password strategy's `password/2` function called `Ash.Igniter.codegen("add_password_authentication")` internally, and then `igniter/0` called it again with `"add_password_auth"`. 

Since `Ash.Igniter.codegen/2` appends `_and_<name>` when a task already exists, this produced the duplicated migration name `add_password_authentication_and_add_password_auth`.

Extracts `AshAuthentication.Igniter.codegen_for_strategy/2` to centralize migration naming with a consistent `add_<strategy>_auth_strategy` pattern across all strategies. Adds test assertions for migration names.


jimsynz (approved):
duplicate

Issue #1155, opened 2026-04-24
Remote IP in context is proxy IP not user IP


quintinm-dev:
Hello, thank you for all the amazing work that goes into Ash!

In some of my Ash actions I have been reaching into the `context.ash_authentication_request.remote_ip` as a key for `AshRateLimiter`. But I am observing that the addresses are from my proxy instead of the actual users, and this matches the dispatcher taking the `remote_ip` from `peer_data`
https://github.com/team-alembic/ash_authentication/blob/62e5a0651d97c545aa66b0afba6d5cc807f4bde0/lib/ash_authentication/plug/dispatcher.ex#L80

Beyond my specific rate limiting issues I think this affects the logs written by `AshAuthentication.AddOn.AuditLog`.

I have a workaround but I'm wondering if there's appetite to fix this more generally. I'm already using `RemoteIp` https://github.com/ajvondrak/remote_ip to rewrite the conn's remote IP for logging, so maybe we could add a configuration option to take from `conn.remote_ip` directly instead?

---

Aside: how bad is it to be writing `ash_authentication_context` into my Ash action context myself? I started rate limiting by IP for actions that aren't authentication related, and it seemed convenient to have a single read path for getting the remote IP. But it feels bad to be setting authentication context for actions that are distinctly not authentication related 😅

quintinm-dev:
Ah interesting I see https://github.com/team-alembic/ash_authentication/pull/1154
feel like this would be an issue there too ;)

jimsynz:
fixed!

PR #1156, opened 2026-04-25
fix: use conn remote ip over peer data

quintinm-dev:
Hello! Thanks again for all the great work here.

This is my proposed fix for https://github.com/team-alembic/ash_authentication/issues/1155. It gives those of us running `RemoteIp` the option to use the `conn.remote_ip` directly so that we record actual client IP addresses in audit logs.

### DIFF
--- a/documentation/dsls/DSL-AshAuthentication.AddOn.AuditLog.md
+++ b/documentation/dsls/DSL-AshAuthentication.AddOn.AuditLog.md
@@ -27,6 +27,20 @@ defmodule MyApp.Accounts.User do
 end
 ```
 
+## Request remote IP configuration
+
+Audit log request metadata stores `remote_ip` from
+`Plug.Conn.get_peer_data(conn).address` by default.
+
+If your app is behind trusted proxy middleware that has already normalized
+`conn.remote_ip`, you can opt in to using that value instead:
+
+`config :ash_authentication, request_remote_ip_source: :conn`
+
+Supported values are `:peer_data` (default) and `:conn`. Only use `:conn` when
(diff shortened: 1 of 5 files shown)


jimsynz:
Welcome @quintinm-dev 👋 

Seems like I entirely missed the `conn.remote_ip` option while working on it. why don't we just change the plug to use `conn.remote_ip || get_peer_ip(conn)` ?

quintinm-dev:
Thanks @jimsynz :wave:

> `conn.remote_ip || get_peer_ip(conn)`

Ah yeah that's probably better. 

I had some vague recollection of reading this:
https://hexdocs.pm/remote_ip/algorithm.html#:~:text=Notice,address,-%2E
so I was worried about forcing people onto the `conn.remote_ip` if they were used to `peer_data`. But on deeper thought:
1) that warning is specific to `RemoteIp` so the only surface there would be people that are already misusing RemoteIp. That's not really AshAuthentication's problem to solve.
2) the Plug docs explicitly recommend remote_ip for this, and it defaults to the peer's IP anyway https://hexdocs.pm/plug/1.19.1/Plug.Conn.html#:~:text=It%20defaults%20to%20peer%27s%20IP

So maybe we can do `conn.remote_ip` on its own? A quick look at the latest versions of the plug adapters does support this:
- https://github.com/mtrudel/bandit/blob/c70e175b68e372bdbd7eb2bf81d236de0ef831b0/lib/bandit/pipeline.ex#L81
- https://github.com/elixir-plug/plug_cowboy/blob/fd0841f4ef974b95a5050a2b5959585c56fb3776/lib/plug/cowboy/conn.ex#L15

PR #1162, opened 2026-05-02
fix: respect optional wax_ dependency for WebAuthn

jechol:
## Summary

Fixes compilation of downstream apps that depend on `ash_authentication` without `:wax_`, even when they do not use WebAuthn.

`wax_` remains required for WebAuthn itself. The real WebAuthn `Actions`/`Plug` implementations, Wax struct usage, and CBOR calls now compile only when the optional dependency is available. Configuring WebAuthn without `:wax_` now raises a clearer verifier error with the dependency and recompile steps.

## Test plan

- [x] `mix test test/ash_authentication/strategies/webauthn/plug_test.exs test/ash_authentication/strategies/webauthn/actions_test.exs test/ash_authentication/strategies/webauthn/cose_key_test.exs`
- [x] `mix deps.compile ash_authentication --force` in a parent app without `:wax_`
- [x] `mix compile` in the parent app

### DIFF
--- a/lib/ash_authentication/strategies/webauthn/actions.ex
+++ b/lib/ash_authentication/strategies/webauthn/actions.ex
@@ -10,428 +10,469 @@ defmodule AshAuthentication.Strategy.WebAuthn.Actions do
   with Ash to persist users and credentials.
   """
 
-  alias Ash.{Changeset, Query}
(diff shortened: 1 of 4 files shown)


jimsynz (approved):
conditional

PR #1163, opened 2026-05-03
feat: add Okta authentication strategy

zachdaniel:
Adds `AshAuthentication.Strategy.Okta`, a thin OIDC-based wrapper for Okta Workforce/CIAM. Covers the same surface as the existing Auth0, Microsoft, and Google strategies: a DSL block, an Igniter generator (`mix ash_authentication.add_strategy okta`), a tutorial, and example wiring in the test fixtures.

Notable choices:

- Defaults to `authorization_params: [scope: "profile email"]`; Assent's OIDC strategy prepends `openid` automatically.
- Documents the Org vs Custom Authorization Server distinction in both the moduledoc and the tutorial — the most common Okta misconfiguration.
- The `groups` claim is documented but not normalised; user/group sync is left to a separate SCIM-based solution.

Also closes a latent gap in the OIDC verifier: `prevent_hijacking/2` was private to `OAuth2.Verifier` and the OIDC verifier never called it, so OIDC-derived strategies (Microsoft, Google, and now Okta) silently skipped the OAuth + password account-hijacking guard. The function is now public and called from `Oidc.Verifier.verify/2`.


zachdaniel:
Will follow up with a new logo in AshAuthenticationPhoenix once this is merged. @jimsynz if you have a sec to look it over. The main thing to notice is that we weren't verifying the hijacking protection for oidc wrappers before but now we are.

jimsynz (approved):
Chef's kiss

PR #1165, opened 2026-05-04
Oauth2 server

zachdaniel:
So: claude helped me quite a bit with this one and it's in a sensitive area. It needs thorough review, but I want to open the PR for visibility as I know this is something that a lot of people are looking for (in many cases because they specifically want an authenticated MCP server).

I myself have reviewed the shape but will be doing code review on this PR as well and refining it.


jimsynz:
Did a careful read of the OAuth2 server logic. Here are my findings — happy to be wrong on any of these, I've called out what I verified vs. what's speculative.

## Looks good
- PKCE is **required**, not optional — `Authorize.validate_request/2` enforces `code_challenge_method == "S256"` (`authorize.ex:52`), `code_challenge` is `allow_nil?: false` on the resource (`resources.ex:77`), and `Token.verify_pkce/2` only accepts requests with a `code_verifier` (`token.ex:106`, with a catch-all `_, _` falling through to `{:error, :pkce}`).
- Codes are **single-use** via the `:consume` action + `consumed_at` attribute (`token.ex:80-91`).
- Codes are **TTL-checked** via `check_not_expired/1` (`token.ex:104-108`).
- Codes are **bound to client_id**, **redirect_uri**, **resource_uri** via `check_client_match`, `check_redirect_match`, `check_resource_match` before exchange.
- PKCE comparison uses `Plug.Crypto.secure_compare` (timing-safe).
- JWT `aud` is correctly set to `server.resource_url()` per RFC 9068 §2.2 (not `client_id` — `client_id` is a separate claim), and verify enforces `iss`/`aud`/`exp` (`jwt.ex:67-100`).
- HS256 is pinned via `@signer_alg "HS256"` — no `alg: none` foot-gun.
- Refresh tokens are stored hashed (`token_hash` attribute + `by_token_hash` identity).
- Consent privilege-escalation is prevented (`Authorize.consented?` only returns true when stored scope ⊇ requested scope — `authorize.ex:87`).
- DCR validates `redirect_uris` to https-only (or http on localhost/loopback) with no fragment (`register.ex:46-58`).
- `Cache-Control: no-store` on token/metadata responses (in the AAP layer).

## Real concerns

### HIGH

**Authorization codes stored in plaintext as the UUID primary key.** `token.ex:73` looks codes up via `Ash.get(authorization_code_resource, code_id)` — the code-value sent to the client IS the row's `id`. Refresh tokens take the hashed approach (`resources.ex:125,170`) and that pattern should be applied here too — DB leak ≠ instant replay. UUIDv4 entropy is high but defence in depth matters.

### MEDIUM

**Redirect URI matching is plain `==` / `in`.** `authorize.ex:155` (`uri in uris`) and `token.ex:139` (`uri == code_uri`). RFC 8252 §7.3 talks about byte-exact match *after* normalization (scheme/host case, default-port elision, trailing-slash). Today `https://app.example` and `https://app.example/` are not equal; equally `HTTPS://APP.EXAMPLE/` won't match a stored lowercase form. Document the exact-match contract or normalize before comparing.

**Dynamic client registration is intentionally unauthenticated** (the moduledoc says "v1 supports public clients only"). That's defensible, but I'd:
- Add a `require_initial_access_token` option per RFC 7591 §3 (default off), so operators who want gated registration can switch it on without modifying the strategy
- Make the v1 trust model louder in the public moduledocs (anyone reachable can register clients with arbitrary `redirect_uris` and immediately participate in flows; current rate-limit / abuse story = none)

**No "max scope" allowlist enforcement on register or authorize.** A DCR client can request any scope strings. `server.scopes()` is joined into the response (`register.ex:91`) but I don't see validation that requested scopes at authorize-time are a subset of `server.scopes()`. If you want to enforce a fixed catalogue, this needs a check in `validate_request/2`.

**No `Pragma: no-cache` companion to `Cache-Control: no-store`** on token responses (RFC 6749 §5.1 calls for both). Cache-Control alone is sufficient on modern HTTP, so this is cosmetic, but the spec compliance scanners will notice.

### LOW / cosmetic
- `register.ex` builds `attrs` from `Map.get(params, ..., default)` — fine, but worth a quick re-read for any missing `is_binary/1` guards on user-controlled string fields that hit a changeset.
- `:invalid_target` error messaging in `resolve_resource/2` could be clearer about what URL was expected vs. received (don't echo the user-supplied URL back unescaped though — `Errors.describe_token_error/1` handles that on the AAP side).

## Style nit (non-blocking)
There are 4 commits on this branch with no Co-Authored-By tag and lowercase messages without a conventional-commits prefix — the project's git history uses `feat:` / `improvement:` / `chore:` etc. Worth squashing or amending before merge.

---

Headline: **the security-critical parts (PKCE, single-use codes, TTL, audience binding, JWT alg pinning) are all in place and correct.** The HIGH item is defence-in-depth (hash the codes) and the MEDIUM items are about hardening + documenting an intentional v1 limitation. Not as alarming as your "claude helped me, needs thorough review" comment suggested — Claude did a good job here.

Companion review on the AAP side: team-alembic/ash_authentication_phoenix#737.

zachdaniel:
Working through the other things, but for this one:

> DB leak ≠ instant replay

To my understanding the fact that we require PKCE explicitly makes this not a problem, because the client has a challenge that the server never sees. You can't replay with just the db row and so hashing adds no functional security benefits.

zachdaniel:
Looks like some issues from the webauthn work unrelated to these changes.

@jimsynz this should be ready for re-review, will get started on AshAuthenticationPhoenix now.

lib/ash_authentication/oauth2_server.ex
```diff
@@ -0,0 +1,364 @@
...
+  | `:sign_in_path` | `nil` | Path to redirect unauthenticated `/oauth/authorize` requests to. When `nil`, returns 401. |
+  | `:initial_access_token` | `nil` | When set, `POST /oauth/register` requires the request to present a matching `Authorization: Bearer …` token (RFC 7591 §3). When `nil` (default), dynamic client registration is open — see the trust-model note below. |
+
+  ## Dynamic client registration
+
+  By default, `POST /oauth/register` is open — any client can register
+  itself with its desired `redirect_uris`. This is the standard mode for
+  OAuth dynamic client registration (RFC 7591) and matches what MCP and
+  most similar flows expect; user-facing protection lives further down
+  in the consent screen and the audience-bound access tokens.
```
maennchen: Can we make this opt-in instead of opt-out?

lib/ash_authentication/oauth2_server/token.ex
```diff
@@ -0,0 +1,334 @@
...
+           |> Ash.update(authorize?: false),
+         {:ok, access_token, claims} <-
+           Jwt.mint(server,
+             sub: row.user_id,
+             client_id: row.client_id,
+             scope: row.scope
+           ) do
+      {:ok, access_token, raw, claims}
+    end
+  end
```
maennchen: `check_refresh_validity` is not done atomically. This could lead to two new tokens being issued with the right timing.

Can we make the `rotate` action fail in case `rotated_to_id` is not nil in the DB?

Or lock the row for update in a transaction so that only one transaction can run at a time.

lib/ash_authentication/oauth2_server.ex
```diff
@@ -0,0 +1,364 @@
...
+    |> String.trim_trailing("/")
+  end
+
+  defp normalize_path(nil), do: nil
+  defp normalize_path("/"), do: nil
+  defp normalize_path(path), do: path
+
+  defp normalize_port("http", 80), do: nil
+  defp normalize_port("https", 443), do: nil
+  defp normalize_port(_, port), do: port
```
maennchen, line 407: RFC 9700 §4.1 requires exact string matching. The current normalization collapses paths that differ only by a trailing slash, which is a deviation.

lib/ash_authentication/oauth2_server/authorize.ex
```diff
@@ -0,0 +1,218 @@
...
+    with :ok <- require_eq(params, "response_type", "code", "unsupported_response_type"),
+         {:ok, client} <- load_client(server, params),
+         :ok <- check_redirect_uri(params, client),
+         :ok <- require_eq(params, "code_challenge_method", "S256", "invalid_request"),
+         {:ok, resource} <- resolve_resource(server, params),
+         {:ok, code_challenge} <- require_present(params, "code_challenge"),
+         {:ok, scope} <- require_present(params, "scope"),
+         :ok <- check_scopes(server, scope),
+         {:ok, redirect_uri} <- require_present(params, "redirect_uri"),
+         {:ok, state} <- require_present(params, "state") do
```
maennchen, line 77: Can we document somewhere that clients should use a cryptographically random state value. (Often people stick things like redirect urls into state which is not safe.)

lib/ash_authentication/oauth2_server/jwt.ex
```diff
@@ -0,0 +1,115 @@
...
+      hardcoded to a version constraint and is not customizable per-token.
+    * **Hot-path verify** — the resource server validates a token on every
+      protected request. Verify here is signature + claims only, no user
+      load. The bearer plug controls when the user record is fetched.
+    * **Decoupling** — these tokens identify a user by their primary key
+      (`sub`), not via an AshAuthentication strategy, so we don't require
+      the user resource to declare `authentication.tokens.enabled? true`.
+  """
+
+  @signer_alg "HS256"
```
maennchen, line 28: I would prefer to offer `RS256` / `ES256` as well so that we can work with asymmetric key pairs as well. (zero trust)

lib/ash_authentication/oauth2_server/register.ex
```diff
@@ -0,0 +1,160 @@
...
+        :ok
+
+      expected when is_binary(expected) ->
+        presented = Keyword.get(opts, :initial_access_token)
+
+        if is_binary(presented) and Plug.Crypto.secure_compare(expected, presented),
+          do: :ok,
+          else:
+            {:error, "invalid_client_metadata",
+             "registration requires a valid initial access token"}
```
maennchen: According to RFC 7591 §3.2.2, this should not produce a 400 since this is not a validation error of the client.

> When an OAuth 2.0 error condition occurs, such as the client
   presenting an invalid initial access token, the authorization server
   returns an error response appropriate to the OAuth 2.0 token type.

Since the initial access token is usually a Bearer token, I would expect a 401 with an appropriate `WWW-Authenticate` header.

lib/ash_authentication/oauth2_server/register.ex
```diff
@@ -0,0 +1,160 @@
...
+
+    * `:initial_access_token` — the bearer token the request presented
+      (or `nil`). When the server has `:initial_access_token` configured,
+      this MUST match (constant-time) or registration is rejected.
+
+  Returns `{:ok, client_record, response_body}` on success or
+  `{:error, code, description}` on a validation failure suitable for a
+  400 DCR error response.
+  """
+  @spec register(server :: module(), params :: map(), opts :: keyword()) ::
```
maennchen, line 45: I would prefer to enforce some kind of rate limiting here.

Also recommended in RFC 7591 §3:

> These requests MAY be rate-limited or otherwise
   limited to prevent a denial-of-service attack on the client
   registration endpoint.

lib/ash_authentication/oauth2_server/jwt.ex
```diff
@@ -0,0 +1,115 @@
...
+      aud == expected -> :ok
+      is_list(aud) and expected in aud -> :ok
+      true -> {:error, :invalid_audience}
+    end
+  end
+
+  defp check_aud(_, _), do: {:error, :invalid_audience}
+
+  defp check_exp(%{"exp" => exp}) when is_integer(exp) do
+    if System.system_time(:second) < exp, do: :ok, else: {:error, :expired}
```
maennchen: I would recommend allowing some configurable clock skew here. I see it often that OAuth2 Servers are slightly out of sync.

RFC 7519 §4.1.4

> Implementers MAY provide for some small leeway, usually no more than
   a few minutes, to account for clock skew.

The same applies to other time based claims like nbf.

maennchen (commented):
This is already pretty good from a security view. But I have a few recommendations.

lib/ash_authentication/oauth2_server.ex
zachdaniel: Given how I anticipate the majority of users using this, I think we should default it to on. But the way I'll do that is making it opt-out in the configuration, but the installer will set it to `true` on installation with a comment explaining, so that it is clear in the user's generated code what they are opting into.

lib/ash_authentication/oauth2_server/jwt.ex
zachdaniel: Plan is to follow up with this, as it is a non-trivial addition.

lib/ash_authentication/oauth2_server/register.ex
zachdaniel: Plan is to follow up with this, as it is a non-trivial addition.

zachdaniel:
@jimsynz I think there may be some kind of issue with deps caching?

zachdaniel:
Okay, for rate limiting, I added a bunch of stuff in this PR but I'm just going to remove it and replace it with a snippet in the docs on how you might do this with a plug. It's too complex to include in its current form in the first pass at this IMHO. (and too easy to just add a plug for the implementor).

lib/ash_authentication/oauth2_server/register.ex
zachdaniel: Actually I've decided to drop this in as a recommended step in the documentation for now. There are plenty of options to rate limit at the router boundary.

zachdaniel:
We are transplanting this code over here:

https://github.com/ash-project/ash_authentication_oauth2_server

PR #1170, opened 2026-05-18
fix: propagate Ash context opts in TOTP setup confirmation changes

arjan:
Ensure Jwt.verify, token lookups, revocation, and pending-setup JWT/store_token receive tenant and other execution opts so multitenant token resources work.


arjan:
Not sure what I'm supposed to do about these CI things 🤷🏻‍♂️

jimsynz (approved):
context is for kings

PR #1174, opened 2026-05-30
feat: check email_verified on google strategy

### LINKED ISSUE #1173: No guard for account hijacking via email_verified: false accounts on google
I had to add an ash change to protect my app from allow non verified google accounts to sign in to existing confirmed accounts, basically hijacking or stealing an account they actually have no right to.

```elixir
defmodule MyApp.Accounts.Changes.RejectUnverifiedGoogleLink do
  @moduledoc """
  Blocks a Google sign-in from resolving to a pre-existing account when Google
  reports the email as **unverified**.

  `:register_with_google` upserts on the shared `:unique_email` identity, so a
  Google sign-in for an address that already has an account signs the caller
  into that account. AshAuthentication's `prevent_hijacking?` guard only refuses
  this when the existing row is *unconfirmed* — a confirmed row (registered via
  password, magic link, or a prior verified Google login) is signed into
  regardless of whether Google verified the email.

  A Google (Workspace) account can present an arbitrary email with
  `email_verified: false`, so without this guard an attacker controlling such an
  account could obtain a session on a victim's confirmed account. We therefore
  reject the sign-in whenever the email is unverified and an account already
  exists.

  When no account exists yet we let the unverified sign-in through: it creates a
  new *unconfirmed* account (see `MyApp.Accounts.Changes.SetGoogleUserDetails`),
  which the `:confirm_new_user` add-on then gates behind the normal
  email-ownership confirmation.
  """
  use Ash.Resource.Change
  require Ash.Query

  @impl true
  def change(changeset, _opts, _context) do
    Ash.Changeset.before_action(changeset, &reject_if_linking_unverified/1)
  end

  defp reject_if_linking_unverified(changeset) do
    user_info = Ash.Changeset.get_argument(changeset, :user_info) || %{}
    email = user_info["email"]

    if verified?(user_info["email_verified"]) or is_nil(email) or
         not account_exists?(changeset, email) do
      changeset
    else
      Ash.Changeset.add_error(changeset,
        field: :email,
        message: "is already registered. Sign in with your original method to link Google."
      )
    end
  end

  .......
end

```

jarlah:
fixes #1173 

Defensive fix, that only targets google at the moment, and its possible to optin for other strategies that reliably supports it.

This is pair programmed with claude code, but more to a level of hand holding a stupid junior, because i used sonnet model and it really wasnt that smart. But it helped me figure out how things worked.


lib/ash_authentication/strategies/google/dsl.ex
```diff
@@ -30,7 +30,7 @@ defmodule AshAuthentication.Strategy.Google.Dsl do
 
       #{strategy_override_docs(Google)}
       """,
-      auto_set_fields: [icon: :google, assent_strategy: Google]
+      auto_set_fields: [icon: :google, assent_strategy: Google, require_email_verified?: true]
```
jarlah, line 33: verify if this correct and/or add actual test for the strategy ... yeah i think ill look into that. Well its tests for the actions ...

zachdaniel (approved):

jarlah:
@zachdaniel will this be merged or discussed since it still sitting here?

zachdaniel:
@jarlah yes we are discussing/working on something related to this. Things take time 😄

jarlah:
no problem :) just wanted to know why it stalled. Im completely fine with if it gets rejected for a better or alternative solution. No fomo on merges 🤣

jarlah:
@zachdaniel this can be closed now ? it takes a wildly different and immature path than the solution you added for the CVE :) already using rc.11 in production.. living on the edge :P

zachdaniel:
Yes 😄 Thank you for your contribution!

Issue #1171, opened 2026-05-27
Support `:private_key` in `:apple` strategy


schwarz:
Currently only `:private_key_path` is supported, but `assent` also lets you [set the `:private_key` directly](https://hexdocs.pm/assent/Assent.Strategy.Apple.html#:~:text=or%20alternatively%20set%20%3Aprivate_key%20with%20the%20content%20of%20the%20private%20key.).
This can be preferable in certain deployment settings.

`:private_key_path` is currently marked as required, but `assent` simply uses `:private_key` if `:private_key_path` is not present. Instead of either being required, I think it should be validated that exactly one of them is given. 
If there's no objection to this I'll put together a PR!

jimsynz:
Implemented in #1177 (with a follow-up in #1180). The `:apple` strategy now accepts `private_key` directly, and the verifier requires exactly one of `private_key` or `private_key_path` via `validate_exclusive_secret/2`. Thanks @schwarz!

PR #1177, opened 2026-06-05
improvement: support `private_key` in Apple strategy

schwarz:
Getting the ball rolling on #1171 

Adding `private_key` in `oauth2/plug.ex` was necessary for the callback to complete, but I'm not sure about it.
Assent prefers `private_key_path` so that's the hangup we report in `verifier.ex` if neither is present.


jimsynz (approved):
keys!

PR #1178, opened 2026-06-14
feat(WebAuthn): added support for Passkey-only mode without a user identity field

simpers:
# Goal

Make the WebAuthn strategy production-grade: unlock **passkey-first authentication** (no identity column on the user resource), close the spec-completeness gaps against WebAuthn Level 2/3, harden the ceremony handling, and decouple the strategy from its crypto backend.

The full status against the spec is documented in the new [WebAuthn tutorial](documentation/tutorials/webauthn.md), which ships a feature matrix and a guide to the supported configuration modes (identity + passkey, passkey-first, second-factor-only, multi-tenant).

# 1. Passkey-only mode (no identity field)

New **required** `require_identity?` DSL option (no default — the developer chooses the mode per resource):

- `true` — classic flow: user resource exposes an `identity_field`
  (default `:email`); credentials are offered via `allowCredentials`.
- `false` — passkey-first: the user resource needs **no identity column**.
  Users sign in via a discoverable credential; the credential id resolves
  the user at verification time. Pairs with `resident_key :required`.

The runtime already supported this; the option relaxes the compile-time checks (transformer, register/sign-in action shapes, `SignInPreparation`).

# 2. `AshAuthentication.WebAuthnCredential` extension

New Spark extension for credential resources. Scaffolds all required attributes, the `belongs_to` user relationship, the unique credential-id identity, and the CRUD actions — every name configurable, everything validated at compile time. Replaces the boilerplate the docs previously told you to copy-paste.

Scaffolded attributes now cover the full spec surface: `credential_id`, `public_key` (COSE/CBOR Ash type), `sign_count`, `user_handle`, `transports`, `backup_eligible`, `backed_up`, `discoverable`, `label`, `last_used_at`.

# 3. Spec-complete registration options

The registration challenge response now carries everything `navigator.credentials.create()` requires:

- **`user` descriptor** — spec-compliant opaque handle (random ≤64 bytes,
  no PII) for new users; a stable, shared handle when adding credentials to
  an existing user. Persisted on the credential (`user_handle`) so all of a
  user's passkeys present one account identity. In passkey-first mode the
  client can pass `display_name` to label the account inside the passkey.
- **`pubKeyCredParams`** — the algorithms the backend actually verifies
  (ES256, EdDSA, ES384/512, PS256/384/512, RS256/384/512), in preference
  order.
- **`excludeCredentials`** — prevents duplicate enrollment of the same
  authenticator (actor's credentials on `add_credential`; identity-matched
  credentials on registration in identity mode).
- **`credProps` extension** — requested at registration; the
  client-reported `rk` result is stored as `discoverable`.

# 4. Richer credential records

- **`transports`** (from the client's `getTransports()`, sanitized against
  the spec-registered set) — echoed back as `allowCredentials` hints so
  browsers route straight to the right authenticator.
- **Backup flags (BE/BS)** — `backup_eligible` and `backed_up` from the
  authenticator data; `backed_up` is refreshed on every assertion (a
  credential can become synced after registration).

# 5. Security hardening

- **Sign counter enforcement (clone detection, §6.1.1)** — new
  `sign_count_policy` option: `:reject` (default), `:log` (allow but keep
  the stored high-water mark so a clone keeps tripping the check), or
  `:ignore`. Synced passkeys (constant-zero counters) are never flagged.
- **Per-ceremony challenge storage** — session challenges are keyed by
  ceremony type and strategy name, so a registration and a sign-in in two
  tabs no longer clobber each other, and a sign-in can never consume a
  registration challenge.
- **Attestation trust verification** — all four conveyances (`none`,
  `indirect`, `direct`, `enterprise`), plus `trusted_attestation_types` and
  `verify_trust_root?` options (permissive defaults preserve the previous
  behavior). FIDO MDS metadata loading is delegated to `wax_` app config
  and documented in the tutorial.

# 6. Backend adapter

All cryptographic ceremony work goes through a new `AshAuthentication.Strategy.WebAuthn.Adapter` behaviour — challenge generation, verification, and challenge (de)serialization for session storage. Challenges are opaque terms; verification returns normalized result maps. The default `Adapters.Wax` is behavior-identical (the entire pre-existing test suite passed unchanged), and a custom adapter is one DSL option away — this is the escape hatch for future needs like Related Origin Requests.

# 7. Custom registration fields with `secret?` confirmation

`register_action_accept` accepts `{field, secret?: boolean}` tuples so UI layers can render appropriately-typed inputs; attributes marked `sensitive?: true` must declare `secret?` explicitly, enforced at compile time.

# Documentation

- New **WebAuthn tutorial**: setup, the four configuration modes, ceremony
  phases, attestation guide (incl. FIDO metadata), a maintained feature
  matrix, and a migration warning for legacy **U2F** deployments (the
  `appid` bridge extension is deliberately not implemented; U2F-era keys
  must be re-registered).
- The `WebAuthnCredential` DSL reference is now actually published (it was
  generated but never registered in the docs config).

# Tests

Full coverage for every feature above: plug round trips (challenge → register against the server-issued challenge), clone-detection scenarios, synced-passkey (zero-counter) flows, restrictive attestation rejection, session-scoping isolation, passkey-only register→sign-in cycles, adapter swappability via a stub backend, and multi-passkey enrollment. Fixtures gained `flags:` and `challenge_bytes:` options to simulate synced passkeys and to answer real server challenges.

# Breaking changes / migration notes

- `require_identity?` is required on every `webauthn do ... end` block.
  Existing resources add `require_identity? true` — one line.
- Assertions with non-increasing sign counters are now **rejected by
  default** (`sign_count_policy :reject`). Set `:log` or `:ignore` to keep
  the old permissive behavior.
- The credential resource gains nullable columns (`user_handle`,
  `transports`, `backup_eligible`, `backed_up`, `discoverable`) — run
  `mix ash.codegen <name>` to pick up the migration. Existing rows stay
  valid.


jimsynz (changes requested):
## Review — WebAuthn passkey-only mode

Thorough, well-documented work: the ceremony-security refactor (per-ceremony/per-strategy challenge session keys, single-use on failure, no unsafe deserialisation) is a genuine improvement over `main`, the `require_identity?` relaxation is implemented without weakening identity mode, and the new extension reuses the existing `Utils`/`Validations` scaffolding rather than inventing a parallel one. Test coverage is broad and the tutorial is largely accurate.

It's not mergeable as-is: CI is red, there's a new pre-auth information disclosure, and — following discussion with the maintainer — passkey-only mode needs to be gated so it can't be reached by default or combined with other strategies. Details below, blockers first.

---

### Blocking

**B1. CI: dialyzer fails (4 errors) — a too-narrow `@spec`, not a runtime bug.**
All four errors trace to one stale contract. `AshAuthentication.Utils.maybe_build_attribute/4` declares its type argument as `atom | module`, but `webauthn_credential/transformer.ex:105` is the first call in the codebase to pass a composite type — `{:array, :string}` for `transports`. That tuple violates the contract, so dialyzer marks the call "will not succeed" and, because the `with` chain has no `else`, treats everything after it as dead (hence `maybe_build_primary_key/1`, `maybe_build_unique_identity/2`, `validate_unique_identity/2` reported as "never called"). Runtime is fine — Ash accepts `{:array, :string}`, as the migration and snapshots show. **Fix:** widen the spec in `utils.ex`, e.g. `Ash.Type.t() | {:array, Ash.Type.t()}`, then re-run dialyzer: lines 105→end were never actually type-checked, so a new warning may surface — the most likely being the non-exhaustive `case` in `validate_unique_identity/2` (see B6).

**B2. CI: `mix spark.formatter --check` also red.**
The shown run predates the current head; `.formatter.exs` looks complete at head (all 17 `webauthn_credential` options present), so please re-run to confirm it's stale. (The `audit` failure is `mint 1.9.1`, a main-level advisory — this PR's `mix.lock` is identical to `main` — so it isn't introduced here and will clear on a rebase; no action needed in this PR.)

**B3. Security: new unauthenticated credential-ID enumeration oracle** (`plug.ex`, `registration_exclude_ids`).
`registration_challenge` is public and unauthenticated. In identity mode (the default) it now looks up the submitted identity and returns that user's real `excludeCredentials` (base64url credential IDs) when the account exists, or `[]` when it doesn't. This is new — pre-PR the endpoint did no lookup. The inline comment says it "doesn't leak account existence beyond what the register action already reveals," but it leaks *credential IDs*, which is strictly more than an existence bit. `excludeCredentials` is a legitimate anti-re-enrolment feature, but exposing it pre-authentication is a degradation. Gate it behind a verified identity/OTP step, or drop the pre-auth lookup and accept the double-registration UX. At minimum, correct the comment — it overstates the safety. Related: these challenge endpoints have no rate limiting and aren't covered by the audit-log add-on (challenge generation isn't an Ash action), so there's no throttling or after-the-fact visibility on this abuse surface.

**B4. Passkey-only mode must be disabled by default.**
Change `require_identity?` from `required: true` (no default) to `default: true`. Safe-by-default, and it removes the hard breaking change for existing resources — a bare `webauthn do … end` should compile in identity mode. Passkey-only becomes an explicit `require_identity? false` opt-in.

**B5. Passkey-only mode must require WebAuthn to be the sole strategy.**
Removing the identity field only holds together when nothing else on the resource relies on one — and almost everything does (6 of 7 strategies declare `identity_field`; only `api_key` doesn't, plus confirmation's `monitor_fields`, the magic_link/otp/password sign-in paths, and OAuth upserts all key off it). So `require_identity? false` must be rejected at compile time whenever any other authentication strategy is present. Enforce in `webauthn/verifier.ex` (which already does cross-strategy checks in `validate_hashed_password_optional`):

```elixir
defp validate_passkey_only_is_sole_strategy(%{require_identity?: true}, _dsl_state), do: :ok

defp validate_passkey_only_is_sole_strategy(strategy, dsl_state) do
  case AshAuthentication.Info.authentication_strategies(dsl_state) do
    [%{name: name}] when name == strategy.name ->
      :ok

    _ ->
      {:error,
       DslError.exception(
         path: [:authentication, :strategies, strategy.name],
         message: """
         `require_identity? false` (passkey-only mode) is only supported when WebAuthn is the \
         only authentication strategy on the resource. The identity field is relied on by other \
         strategies and shared authentication machinery; remove the other strategies or set \
         `require_identity? true`.
         """
       )}
  end
end
```

The rule is intentionally "any other strategy," not "any identity-using strategy" — the only combination the narrower rule would permit is passkey-only + api_key, and admitting it would require auditing the whole user model for identity-less tolerance, which isn't worth one edge case. This is consistent with 2FA mode: a second-factor WebAuthn strategy always has a primary strategy present, so it lands on `require_identity? true`, automatically satisfied by the primary's identity field (and WebAuthn's own register/sign-in identity validations are gated off in that mode). Tests to add: `require_identity? false` + any second strategy fails compilation; sole passkey-only strategy compiles; unset defaults to identity mode.

---

### Should-fix

**S1. Sign-in actions are generated even when `sign_in_enabled? false`** (`webauthn/transformer.ex:93–102`).
`register` is gated on `registration_enabled?` and `verify` on `verify_enabled?`, but `build_sign_in_action`/`build_sign_in_with_token_action` are built unconditionally — `sign_in_enabled?` is never referenced in the transformer. Routes/protocol hide them, so it isn't HTTP-exploitable, but the callable Ash actions remain, defeating the documented "verify-only" intent for 2FA-only deployments (and the generator's own `--mode 2fa` sets `sign_in_enabled? false`). Gate them to match Password.

**S2. Non-exhaustive `case` in `validate_unique_identity/2`** (`webauthn_credential/transformer.ex:355`).
Only matches `%{keys: keys}`; `Resource.Info.identity/2` can return `nil`. Runtime-safe today (guarded by the build immediately prior) but fragile, and the prime candidate to newly trip dialyzer once B1 un-hides this code. Add a `nil ->` clause returning a `DslError`.

**S3. `handle_sign_count`/`update_assertion_state` hardcode attribute names** (`actions.ex:672, 684–688`).
The re-read filter uses `credential_id == ^credential_id` and writes `%{sign_count:, last_used_at:, backed_up:}` literally, while the rest of the module correctly uses `strategy.credential_id_field`, `strategy.sign_count_field`, etc. If any is customised, the update silently matches nothing — the sign count never advances and clone detection degrades to a no-op. The `sign_count`/`last_used_at` hardcoding is pre-existing; this PR extends the pattern with `backed_up`.

---

### Minor / nits

- **Password typespec overreach** (`password.ex:152`): `register_action_accept` now advertises `{atom, [secret?: boolean]}` tuples, but Password's DSL schema (`{:list, :atom}`), transformer, and verifier weren't updated — the tuple form is rejected at compile time. Only WebAuthn supports it. Revert the Password typespec or finish it there.
- **`sign_count_policy :ignore` is untested** — `:reject` and `:log` are covered; the third value (which overwrites/lowers the stored count, distinct from `:log`) has no test.
- **No tampered/expired-challenge tests** — wrong-ceremony reuse and sign-count regression under `:reject` are covered; challenge tamper/expiry are not.
- **Adapter swappability tested shallowly** — the stub's `verify_authentication` and any non-Wax success path are never exercised.
- **Broken assertion via operator precedence** (`plug_test.exs:~66`): `assert session["user_handle"] || session[:user_handle] == body["user"]["id"]` parses as `x || (y == z)`; the truthy binary short-circuits so the equality never runs. Low impact (the property is checked in the round-trip test).
- **`attestation "direct"` footgun**: with default `trusted_attestation_types`/`verify_trust_root? false`, requesting direct attestation performs no actual trust verification. Matches pre-PR behaviour (no regression) but the DSL doc implies verification that doesn't happen — add a cross-reference.
- **`label_field`/`last_used_at_field` skip `validate_attribute`** — the only two built fields with no post-hoc type/nullability validation if user-declared.
- **`SignInPreparation` returns an unfiltered query** in passkey-only mode (and on `find_strategy` error) — the generated read action is vestigial (real flow resolves via credential id), but returning all rows and swallowing the error is weaker than pre-PR. B5 makes this less reachable but the branch is still worth tightening.
- **Redundant re-read** in `update_assertion_state` — the credential is already loaded during the ceremony, then re-read by id to update (one extra query per sign-in).
- **Tutorial** never states `require_identity?` is mandatory; after B4 it should document that identity mode is the default and passkey-only is opt-in.

---

### Design notes (from maintainer discussion — not change requests)

These came up while reviewing and are recorded for context; the conclusions support the current design.

- **Satellite-resource reuse:** no case to unify `WebAuthnCredential` with `UserIdentity`/`TokenResource`. `TokenResource` has no user relationship; the api_key resource isn't extension-scaffolded at all; and `UserIdentity` (OAuth linkage, composite `[uid, strategy]` identity, upsert on token columns) shares only a `belongs_to`-shape with `WebAuthnCredential` (keypair possession, single-key `credential_id`, plain CRUD). The real shared abstraction — `maybe_build_*` in `Utils` + `Validations` — is already reused by all three transformers. Extracting a common "satellite" base over two members would be premature.
- **Audit log / brute force:** the PR wires nothing itself, but the audit-log add-on hooks generically at the resource level (`include_actions`/`include_strategies` default to `[:*]`), so a resource with the add-on logs WebAuthn register/sign-in automatically. Brute-force protection is opt-in per strategy and WebAuthn doesn't expose it — defensible for the assertion step (challenge + public-key auth isn't guessable), but see B3 on the unauthenticated challenge endpoints, which nothing currently throttles.
- **Multi-device / passkey-only account model:** enrolling additional passkeys requires an authenticated session (`add_credential` needs an actor), which is correct — unauthenticated enrolment would be an account-takeover primitive. New devices bootstrap via cross-device hybrid (QR), another auth method, or synced passkeys. The genuine gap is passkey-only + device-bound authenticators + no hybrid, where new-device bootstrap and account recovery become the app's responsibility; worth spelling out in the tutorial.

---

### Verified correct (load-bearing)

`require_identity?` relaxation never assumes `identity_field` in passkey mode while keeping full validation in identity mode; sign-count high-water mark preserved under `:log`; synced-passkey (zero-counter) exemption matches spec §6.1.1; transports sanitised against the spec set; challenge (de)serialisation uses only `Base.encode64`/plain maps (no `binary_to_term`), single-use on both success and cryptographic-failure paths; per-ceremony/per-strategy session keys close a real pre-PR clobbering weakness; attestation defaults exactly reproduce the previously-hardcoded values; orphan-user cleanup removal is safe (single transactional create with `ManageRelationship`); migration adds only nullable columns; the `WebAuthnCredential` DSL doc is now registered in `mix.exs`; generators handle both modes and scaffold the credential resource.

simpers:
> B5. Passkey-only mode must require WebAuthn to be the sole strategy.
**About this whole Passkey-only exclusivity**

In my book, the strategies that WebAuthn would exclude would be the strictly less safe ones, such as Passwords and Magic Links and so on, but not OAuth2, OIDC, Apple, and whatever else, all of which might have their own Passkeys anyway.

So my intended use of allowing people to register using different integrations and adding Passkeys as the (_probably_) only "native" strategy to authenticate will thus not be permitted.

It also makes it difficult to migrate from say having a password one to then only using Passkeys.

I want clarity more clarity on why it has to be like this, as I do not see it as such a one or the other kind of thing. Also, how does it affect custom strategies down the road? I have plans on providing a custom strategy which integrates with Sweden's BankID system, which is _the_ way to identify a real person in Sweden, as an AshAuthentication strategy, and it is as good as a WebAuthn Passkey (when properly configured) for my purposes.

**The rest**
The rest is good and reasonable, although I'm not a fan of the change to defaulting rather than explicitly choosing after installing, but that's okay 👍🏽 

Working on fixing the rest while I wait for more info on **B5**

jimsynz:
Thanks for waiting, and for pushing back on B5 — the migration point in particular was right and changed the design.

**B4.** Still needs doing, and the reason is compatibility rather than taste: people are already on the 5.0.0-rc line with `webauthn do … end` blocks. `required: true` with no default makes every one of those a compile error on upgrade. Please change it to `default: true`. Passkey-only stays available, just as an explicit opt-in.

**B5.** You've convinced me the blanket rule is too strict, so let's make it an opt-out rather than a wall. Add a `compatible_strategies` DSL option to the WebAuthn strategy, a list of strategy names, defaulting to `[]`. The verifier rejects `require_identity? false` when any strategy on the resource isn't either WebAuthn itself or named in that list. So the strict rule is what you get by default, and your cases become explicit:

```elixir
webauthn do
  require_identity? false
  compatible_strategies [:google, :bankid]
end
```

That gives you OAuth2/OIDC coexistence, a path for BankID when you build it, and a route for migrating an existing password app to passkey-only — while keeping the default safe for someone who doesn't know that six of seven strategies key off an identity field.

Please make the verifier's error message name `compatible_strategies` explicitly. The value of an opt-out is entirely in whether the person who hits the wall discovers it exists.

Tests: `require_identity? false` plus an unlisted strategy fails to compile; plus a listed one compiles; unset defaults to identity mode.

Everything else from the review looks addressed — CI is green apart from a stale `bandit` advisory that a rebase will clear. Once B4 and B5 are in I think this is landable.

simpers:
**B5:**
> That gives you OAuth2/OIDC coexistence, a path for BankID when you build it, and a route for migrating an existing password app to passkey-only — while keeping the default safe for someone who doesn't know that six of seven strategies key off an identity field.

I am confused why `require_identity?: false` would conflict with other strategies. Can you elaborate on that?

The flag is meant only to configure WebAuthn itself to require that upfront or not, which may or may not be something you require in the app, independently of whether you in fact do have one (or multiple identities).

For example, if your app has an identity, "email", that is independent of whether or not you require the WebAuthn strategy to **use and require** that identity as part of signing in, versus just relying on the fact that the WebAuthn.Credential itself on its own will identify the account anyway.

So you can configure it in multiple ways. I have use cases where I do require an email, and some where I don't, and I think I would still **not** want the email to be submitted as part of signing in. The Credential itself is enough. But that doesn't mean that an email is required to create an account in general. The `:require_identity?` is only controlling whether or not to force that field/input or not, and I think it does a double-check there where it verifies that the email matches the account identified by the credential itself (there was an exploit/issue I can't remember the details of, but you shouldn't check this too early either).

Just want to make sure that the interpretation of this **isn't** that the strategy itself absolves the account resource of its requirements of having an identity, because that isn't what it means. It simply means that the use of an identity is required or not. But if an identity will be used, it will verify that it exists on the account, but it otherwise does not affect the account itself.

With that said, I see no compatibility issues with my current understanding of it, but if there is something I have missed, please let me know.

**B4:** I thought I had pushed that already, but apparently not. Will get on it and rebase

simpers:
@jimsynz About **B5**:

Claude's words after checking whether there was more to it that I hadn't seen, and based on the understanding I have it is still the same:

_Went and checked where require_identity? is actually read in the transformer/preparation/actions code — every site is scoped to the actions this one webauthn block generates itself:_

- validate_identity/2 — skips WebAuthn's own compile-time check that identity_field exists and is writable/public/unique. Doesn't touch the attribute or any identity constraint on the resource itself.
- build_register_action_accept/1 — whether WebAuthn's own generated register action accepts identity_field. Other strategies' register actions (e.g. register_with_google) are separate action definitions with their own accept/arguments, untouched.
- build_sign_in_action/2 and the sign-in preparation — whether WebAuthn's own sign_in action takes identity_field as an argument, and how it looks the user up at runtime.

_None of these reach into the resource's attribute definitions, its identities do ... end block, or any other strategy's actions. So require_identity? false cannot, by itself, produce the kind of gap I think you're picturing (e.g. an OAuth2 action's upsert_identity colliding with an account that has no identity value)._

_If a developer also chooses to drop the identity attribute or its uniqueness constraint from the account resource, that would break upsert_identity-style actions — but that's an independent resource-design decision, and it can happen with any combination of strategies, with or without WebAuthn in the mix at all. Gating it behind a WebAuthn-specific compatible_strategies list checks the wrong thing in the wrong place: it can't see the actual hazard (a missing/non-unique identity attribute), so it would either block legitimate coexistence (the case I raised originally) or say nothing when the identity attribute really has been removed._

_If there's an invariant worth protecting here, I think it's resource-level — something that checks every strategy relying on an identity field (via identity_field, upsert_identity, etc.) has a corresponding valid attribute — independent of WebAuthn specifically. Happy to help scope that as a separate verifier if it seems worth doing, but I don't think it belongs on this strategy's DSL._

**If the concern is** that there would be an invariant of whether or not an email **SHOULD** always exist or not, based on the strategies that are available on the resource, that kind of validation would belong on the account resource itself.

The flag in question in this branch is entirely isolated to the strategy itself, and it is referring to WebAuthn's identity, and whether one is required or not, because WebAuthn Credentials themselves do not need an identity, as they are unique compared to how other strategies work. One Credential will always directly identify one account through its relationship, and so the identity itself is not needed, but some do still force this (for what reason, I do not know 😅).

PR #1184, opened 2026-07-02
feat(oauth2): opt-in IdP-initiated login via request-phase restart

pinetops:
Adds an `idp_initiated_login?` option (default `false`) to the OAuth2 strategy. When enabled, a callback that arrives with **no stored session** (an IdP/third-party-initiated login — e.g. a Clever or ClassLink portal launch that omits `state`) is treated as a *trigger* to restart the request phase rather than completed directly.

This is the OpenID Connect Core §4 ("Initiating Login from a Third Party") pattern: the inbound response is discarded and the plug redirects into `request/2`, which mints a fresh `state` that is verified on the ensuing SP-initiated callback. **CSRF `state` verification stays intact for every flow** — the alternative of skipping `state` on an empty session is an attacker-controllable downgrade, since "session absent" is indistinguishable from an SP-initiated flow with a stripped cookie. It's also the handling Clever's own docs prescribe (ignore the Clever-initiated login; restart from your own site with your own state).

### Implementation
- New `idp_initiated_login?` DSL option on the OAuth2 strategy, default `false` (no behaviour change unless opted in).
- The branch is added to `maybe_reflect_or_fail/2` **after** the existing form_post interstitial case, so Sign in with Apple is unaffected, and is gated **GET-only**.
- Default behaviour (fail closed on a stateless callback) is unchanged.

### Tests
Stateless GET fails closed by default; with the flag enabled it 302s into the request phase, stores a fresh `state`, and does **not** complete auth from the inbound code. Full suite green.

### Provenance
Rebased onto current `main` (single commit). This mechanism is **running in production** (u2i/teamology) for Clever→EdLink SSO across ~20 school districts: a global Clever launch lands stateless, bounces through the request phase, and completes `state`-verified — validated end-to-end against live Clever and EdLink.

_(Supersedes the earlier `fix/oauth2-allow-empty-session-params` approach on this PR, which skipped `state` on an empty session; that was the downgrade this replaces.)_


jimsynz:
Thanks for the detailed writeup and spec references. The problem is real, but I think the fix disables the wrong thing — and it now collides with code that landed on `main` after this branch was last updated. Laying out the reasoning:

## 1. Conflict with #1187 (merged after this branch)

#1187 repurposed the empty-session `nil ->` branch of `callback/2` to drive the Apple `form_post` interstitial: a cross-site POST arrives without the `SameSite=Lax` cookie, the session is absent, and `maybe_reflect_or_fail/2` renders a same-origin re-POST to recover it.

Removing the `session_params when is_map(session_params) <-` guard means a `nil`/empty session now matches `session_params <-` unconditionally and the `with` proceeds — so it never reaches the `nil ->` branch. That bypasses the interstitial entirely (Apple `form_post` sign-in breaks) and makes `maybe_reflect_or_fail/2` dead code. This needs rebasing onto current `main`, and the two empty-session cases now have to be told apart.

## 2. Disabling `state` on an empty session is an attacker-controllable downgrade

Keying the `state: false` path on "session is empty" means the callback can't distinguish a genuinely IdP-initiated request from an SP-initiated one where the attacker simply didn't send a session cookie. A login-CSRF / account-linking attack against a normal SP-initiated flow is *also* a GET callback, so "empty session" (or even "empty session + GET") contains both the legitimate case and the attack. Session presence isn't authenticated, so this relaxes CSRF protection for all flows on that strategy, not just IdP-initiated ones. This is why RFC 9700 §4.7 treats IdP-initiated as a distinct, riskier case rather than a supported default.

Transport-level checks (source IP / hostname / `Referer`) don't rescue it: the callback is browser-mediated, so its source is the user's (or attacker's) browser, never the provider. The provider is authenticated at a different layer — the server-to-server token exchange (TLS + `client_secret`) plus `iss`/`aud` validation — none of which depends on `state`.

## 3. Both cited providers already expose a secure route

I checked how Clever and ClassLink actually expose their launches:

**Clever** — Portal-initiated launches carry no `state` ("State parameters are not included in Clever-initiated logins"), redirecting to your primary `redirect_uri` with a `code`. Clever's own guidance for apps that want `state`: *"ignore the incoming Clever-initiated login and restart the authentication flow from your own site so you can include `state` yourself"* — i.e. bounce to SP-initiated.

**ClassLink** — Two portal-configured modes:
- **Dynamic state**: "you'll send the user to your company-generated URL, which will generate the appropriate state and redirect the user to LaunchPad." This *is* the SP-initiated bounce, built into their launch model — the callback arrives with a valid `state`, and no library change is needed.
- **Static state**: a fixed string replayed on every launch. Present but constant and not session-bound, so it provides no real CSRF protection and still wouldn't pass Assent's `verify_state` (nothing stored to compare against).

So `state: false` isn't the right integration for either provider — both hand you a route that keeps `state` intact, and one (Clever) explicitly tells you to take it.

## Suggested direction

If the library should help here, the safer opt-in is **"on a stateless callback, redirect to the request phase"** rather than **"skip `state` verification"** — same per-strategy, default-`false` flag, but it *converts* IdP-initiated into SP-initiated instead of disabling the check:

- **ClassLink dynamic state**: pure configuration (point the tile at the strategy's `request/2` endpoint) — arguably no library change at all.
- **Clever**: the portal always hits `redirect_uri`, so the bounce happens at callback time — detect a stateless callback, discard the `code`, `302` to `request/2`. The user has a live Clever session, so they return immediately with a proper `state`. Full CSRF protection, no protocol weakening.

`state: false` would then only ever be needed for ClassLink *static* state — the weakest of the configurations, and not something a default should steer people toward.

Precedence would need spelling out, since the Clever bounce and #1187's interstitial both trigger on an empty session, e.g.:

1. empty session + POST + not reflected → interstitial (#1187)
2. empty session + GET + strategy opted into IdP-initiated → `302` to request phase
3. empty session, otherwise → fail closed (current behaviour)

Happy to pair on the bounce-based version if that direction sounds right.

### References
- Clever: [OAuth Implementation](https://dev.clever.com/docs/oauth-implementation), [Best Practices & Edge Cases](https://dev.clever.com/docs/best-practices-edge-cases)
- ClassLink: [Requesting Authorization for Access](https://help.classlink.com/s/article/Requesting-Authorization-for-Access), [OAuth2 Workflow](https://help.classlink.com/s/article/OAuth2-Workflow)
- [RFC 9700 §4.7](https://www.rfc-editor.org/rfc/rfc9700.html) (IdP-initiated / CSRF)

pinetops:
Ok, so now we flip an IdP originated flow into an ordinary flow with state if idp_initiated_login? is set. Verified with Clever.

jimsynz:
The bounce-based version looks right — the precedence matches what we discussed, the default stays fail-closed, and the new tests pass locally against current `main`.

One blocker before merge though: **`idp_initiated_login?` is a silent no-op on `dynamic_oidc`**, which is the strategy most likely to want it (data-driven multi-tenant SSO is exactly the Clever/ClassLink shape).

The chain:

1. The DSL schema is inherited (`DynamicOidc.Dsl.dsl()` → `Oidc.dsl()` → OAuth2 schema), and `patch_schema/0` doesn't delete the new key, so the option is accepted by the compiler and advertised in the regenerated `DSL-AshAuthentication.Strategy.DynamicOidc.md`.
2. But unlike the other OAuth2-derived strategies (which keep `target: OAuth2` and got the new struct field for free), `DynamicOidc` sets its own target with its own mirror `defstruct` — and that struct didn't get `idp_initiated_login?`.
3. Spark builds entities with `struct/2`, which silently drops keys the target struct doesn't have, and the plug reads the flag with `Map.get(strategy, :idp_initiated_login?, false)` — so instead of a `KeyError` exposing the mismatch, it quietly evaluates to `false`.

Net effect: `idp_initiated_login? true` in a `dynamic_oidc` block compiles, is documented as supported, and does nothing — a stateless GET callback still fails closed.

Fix is small:

- Add `idp_initiated_login?: false` to the `DynamicOidc` defstruct (alphabetically it sits between `identity_resource` and `name`).
- Change the plug to `strategy.idp_initiated_login?` instead of `Map.get(..., false)` — once both structs have the field the defensive default only serves to hide the next field that drifts out of the mirror struct.

Two minor nits while you're in there:

- The new field is inserted after `registration_enabled?` in the OAuth2 struct, typespec, and DSL schema, breaking the otherwise-alphabetical ordering.
- A third test asserting that a stateless **POST** with the flag enabled still takes the interstitial path (not the restart) would pin the `cond` clause ordering that protects Sign in with Apple.

pinetops:
On `dynamic_oidc`: `idp_initiated_login?` is now **rejected at compile time** (DslError from the DynamicOidc verifier) rather than silently accepted.

Reasoning: the flag restarts the request phase to obtain a verifiable `state`, but `dynamic_oidc`'s request phase resolves its provider config (`base_url`/`client_id`/`client_secret`) from a `connection_id` in the request *path*. An IdP-initiated callback carries no `connection_id`, so the restart can't build an authorize URL — the flag would compile, be documented, and never fire. Better to fail loudly.

My actual flow I'm testing is Clever/Classlink->Ed.link->ash_authentication, so I don't need this capability (ed.link needs a custom strategy). I've made this a public lib:
https://github.com/u2i/ash_authentication_edlink

I'm open to exploring whether we can solve for the connection_id issue with a direct clever connection, but I think this would make sense as a separate PR.

One specific question is whether I've solved for the non-availability of this option for the dynamic_oidc strategy in the right way (I throw a validation error at compile time). Happy to change the strategy if needed.

pinetops:
Just a note to say that updating this PR was a bit noisy, but it's ready for review now.

jimsynz:
Rebased this onto current `main` and added one follow-up commit. Both are on `pinetops/oauth2-idp-initiated-login` in this repository rather than on your branch, and I've opened #1205 from it.

That's not a comment on the work — we simply couldn't push to `u2i/ash_authentication`. It's an organisation-owned fork, and GitHub doesn't honour "allow edits by maintainers" for those. The PR reports `maintainerCanModify: true`, which is misleading; the actual permission is `push: false`. Your commit is unchanged and still authored to you.

The rebase was clean. It picks up the `mix_audit` → `mix hex.audit` swap, v5.0.0-rc.13, and the Expunger `ash_domains` change.

The follow-up adds a loop guard the GET branch was missing. The POST side is bounded by `not reflected?(conn)`; the GET side wasn't, so if the session couldn't be persisted, the restarted flow's callback landed back in the same branch and restarted again until the browser hit its redirect limit. The guard is `not state_param?(conn)` — restart only when the callback carries no `state`. A restarted flow always carries one, so its presence means a request phase already ran and the session should have been there. Fail closed instead of bouncing again.

One behaviour change to flag, since it touches your use case directly: ClassLink's static-state mode replays a fixed `state` on every launch, so those launches now fail closed rather than restarting. Dynamic state points the tile at the request phase and never reaches this branch, and Clever omits `state` on portal-initiated logins, so neither of those is affected. If you have a static-state deployment, it would need the portal switching to dynamic state — tell me if that's a problem and we'll work out the alternative.

Apologies this sat for as long as it did. The approach was agreed back in July and it should have gone in then.

jimsynz:
Closing in favour of #1205, which is this work rebased onto current `main` plus the loop-guard follow-up described above. Your commit is in there unchanged and still authored to you — the only reason it's a separate PR is that we can't push to an organisation-owned fork.

CI is green on #1205: 17 checks passing, including `mix test`, `dialyzer`, `credo --strict`, `sobelow` and `audit`.

Thanks for the work, and for pushing back on the first review — the bounce-based design is better than what I originally suggested.

PR #1186, opened 2026-07-02
fix: filter auto-generated get_by_<identity_field> lookup actions

pinetops:
## Problem

The password and magic_link strategy transformers add an auto-generated `get_by_<identity_field>` read action with `get_by: [identity_field]` and `get?: true`. Ash's `Ash.Resource.Transformers.GetByReadActions` is what turns that `get_by` option into an actual `filter` on the action — but `AshAuthentication.Strategy.Custom.Transformer` declared no ordering relative to it. So `GetByReadActions` can run **before** the strategy transformers add their actions, and never synthesizes the filter.

The action then runs **unfiltered**, and because it is `get?: true` (read via `Ash.read_one`), it raises `Ash.Error.Invalid.MultipleResults` as soon as the resource's table holds more than one row — user lookup breaks in any real database. (It passes in single-row unit tests, which is why it can go unnoticed.)

Confirmed at runtime on `main` before the fix:

```
Example.User.get_by_username: get_by=[:username] filter=nil
```

## Fix

Declare `Custom.Transformer.before?(Ash.Resource.Transformers.GetByReadActions)` so the strategy transformers' `get_by` actions exist before the filter is synthesized. One line, at the root cause; fixes both the password and magic_link lookup actions at once (no per-strategy patching).

After the fix:

```
Example.User.get_by_username: get_by=[:username] filter=<username == arg(:username)>
```

## Test

Adds a regression test to `magic_link_test.exs`: with two users present, the auto-generated lookup action returns the correct single user rather than raising `MultipleResults`. Full suite green (716 tests + 22 doctests).

### DIFF
--- a/lib/ash_authentication/strategies/custom/transformer.ex
+++ b/lib/ash_authentication/strategies/custom/transformer.ex
@@ -25,6 +25,13 @@ defmodule AshAuthentication.Strategy.Custom.Transformer do
   @impl true
   @spec before?(module) :: boolean
   def before?(Resource.Transformers.DefaultAccept), do: true
+  # Strategy transformers add auto-generated `get_by_<identity_field>` read
+  # actions (password, magic_link). Ash's GetByReadActions transformer is what
+  # turns their `get_by:` option into an actual filter — so it must run *after*
+  # the strategy transformers have added those actions, otherwise the actions
+  # stay unfiltered and `Ash.read_one` raises MultipleResults once more than
+  # one row exists.
+  def before?(Ash.Resource.Transformers.GetByReadActions), do: true
   def before?(_), do: false
 
   @doc false
--- a/test/ash_authentication/strategies/magic_link_test.exs
+++ b/test/ash_authentication/strategies/magic_link_test.exs
@@ -13,6 +13,31 @@ defmodule AshAuthentication.Strategy.MagicLinkTest do
 
   doctest MagicLink
 
+  describe "auto-generated lookup action (get_by_<identity_field>)" do
+    test "is scoped by a filter, even with multiple users present" do
+      # The strategy transformers auto-generate a `get_by_<identity_field>`
+      # read action with `get?: true`. Ash's GetByReadActions transformer must
+      # run after them to synthesize the filter from `get_by`; otherwise the
+      # action is unfiltered and `Ash.read_one` raises MultipleResults as soon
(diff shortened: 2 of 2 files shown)


jimsynz:
Thanks for this — the diagnosis is spot on and the fix is in exactly the right place.

I reproduced it on current `main`: `get_by_username` compiles with `get_by=[:username]` but `filter=nil`, and applying the one-liner synthesises the filter (`username == arg(:username)`) with the suite still green, so no transformer cycle.

For anyone reading later, the root cause in one paragraph: `get_by:` doesn't build the filter where the action is declared — it's metadata, and `Ash.Resource.Transformers.GetByReadActions` is a separate compile-time pass that scans the actions *present when it runs* and injects the filter into each one carrying a `get_by`. For a hand-written action that's always fine, because the entity is in the DSL before any transformer runs. But `get_by_<identity_field>` is *added* by a transformer (`Custom.Transformer`), so there's a data dependency between two passes — and neither declared an ordering relative to the other. When the topological sort placed `GetByReadActions` first, it scanned before the action existed, matched nothing, and left the action unfiltered; `Ash.read_one` then raises `MultipleResults` once the table holds more than one row.

So the fix is about **ordering the passes**, not about adding a filter to the action — the action definition is already correct, and pinning the order lets the existing implicit generation see the action. Declaring `before?(GetByReadActions)` on the shared `Custom.Transformer` is the right level: it covers password and magic_link (and any future `get_by`-based action) at the root rather than per-strategy.

One thing worth calling out in the description: this isn't quite "breaks in any real database" universally — otherwise it'd have surfaced everywhere immediately. The tie-break between the two unconstrained transformers depends on the full transformer set on a given resource, which varies per app, so some resources landed on the correct order by luck and others didn't. That makes it a latent, per-resource ordering bug — arguably nastier than a deterministic one — and pinning the constraint is what makes it reliable.

Two small suggestions, neither a blocker:
- The regression test only covers the magic_link lookup. Since the fix is shared, that's enough to guard the ordering, but the password reset path (`get_by_<identity_field>` via `RequestPasswordReset`) is the more security-relevant one and has no equivalent test — worth a mirror test or a comment noting the shared coverage.
- If it's easy to tell, it'd be useful to know what recently shifted the tie-break (an Ash version bump reordering transformers?), since it'd flag whether any other unpinned orderings are now fragile too.

PR #1192, opened 2026-07-24
feat(oauth2): allow a strategy to supply its own http_adapter

pinetops:
## Problem

`AshAuthentication.Strategy.OAuth2.Plug` unconditionally overwrites `config[:http_adapter]` with the `:ash_authentication, :http_adapter` application setting (defaulting to `Finch`):

```elixir
defp add_http_adapter(config) do
  http_adapter =
    Application.get_env(:ash_authentication, :http_adapter, {Finch, supervisor: AshAuthentication.Finch})

  {:ok, Map.put(config, :http_adapter, http_adapter)}
end
```

So a strategy cannot use a custom `Assent.HTTPAdapter`, even though Assent supports one via config. There is also no per-strategy way to configure it — only the global application setting.

## Motivation

Some providers return responses that aren't quite standard, and the clean way to handle that is a small custom `Assent.HTTPAdapter` that adjusts the response at the transport layer *before* Assent's spec-compliant OAuth2 parsing runs — keeping Assent itself unmodified and spec-pure.

Concretely: [Ed.link](https://ed.link) wraps its token and profile responses in a `{"$data": ...}` envelope (which is not RFC 6749 §5.1-compliant). The correct fix is a transport adapter that unwraps it — but there was no way to hand that adapter to the strategy, because the plug clobbers it. (I initially proposed a hook in Assent for this — pow-auth/assent#205 — but that was reasonably rejected)

## Change

- Add an `http_adapter` option to the `oauth2` DSL — a module or `{module, opts}` tuple.
- Have `add_http_adapter/2` prefer `strategy.http_adapter` over the application default.

```elixir
oauth2 :my_provider do
  # ...
  http_adapter MyApp.HTTPAdapter.Unwrap
end
```

## Compatibility

Backward compatible: the option defaults to `nil`, so strategies that don't set it fall back to the application setting exactly as before. All OAuth2/OIDC-derived strategies inherit the option (regenerated cheat sheets included).

Tests: a new plug test asserts a strategy's `http_adapter` reaches the Assent config instead of the app default; the existing OAuth2 suite passes.


jimsynz:
Thanks for the writeup — the Ed.link envelope is a real problem, but I don't think this is the layer to solve it at, and I think the capability you need already exists.

`assent_strategy` is a field on the strategy struct. The generic `:oauth2` entity pins it via `auto_set_fields`, but a custom strategy defines its own `Spark.Dsl.Entity` and can set it to whatever it likes. That's not a workaround — it's how every provider in this library is built. `AshAuthentication.Strategy.Github.Dsl.dsl/0` is the whole pattern:

```elixir
OAuth2.dsl()
|> Map.merge(%{
  name: :github,
  args: [{:optional, :name, :github}],
  auto_set_fields: [icon: :github, assent_strategy: Github]
})
|> Custom.set_defaults(Github.default_config([]))
```

`OAuth2.dsl/0` and `Custom.set_defaults/2` are public API, so an `:edlink` strategy is the same shape with your own `Assent.Strategy` module — one that delegates to `Assent.Strategy.OAuth2` and unwraps `$data` in the token and userinfo responses. You get the whole OAuth2 plug pipeline, `state` verification included, and the non-compliance is contained in the one module that knows about it.

That's a better fit than a transport adapter, because the envelope is a payload-shape problem rather than a transport one — and it avoids putting an `http_adapter` option on the DSL surface of all nine OAuth2-derived strategies to serve one provider.

One thing your PR does expose that's worth fixing separately: `add_http_adapter/1` uses `Map.put`, so it clobbers an `http_adapter` that a custom Assent strategy sets in its own `default_config/1`. Anyone needing genuinely transport-level treatment — client certs, a proxy, per-provider TLS — is blocked by that one line, and `Map.put_new` would unblock them without any new DSL. If you'd like to send that instead I'd take it.

I'll leave this open rather than closing it — if the envelope turns out to reach further than the token and userinfo responses, that changes the picture and I'd want to hear it.

pinetops:
I think we can close. I think you're right that doing this via transport (while it helps reduce some duplication) is ultimately not the right solution.

PR #1193, opened 2026-07-29
fix: remove a duplicate call in the installer

vonagam:
Two nothing burgers:

- Noticed that there is a duplicate call to set `[:authentication, :tokens, :token_resource]` in the installer.

- Noticed that in manual installation documentation `require_token_presence_for_authentication? true` is missing even though Igniter does set it and it is documented to be required by `LogOutEverywhere` add-on.


jimsynz (approved):
seeing double

PR #1198, opened 2026-08-20
fix: don't delete the session when a resource requires a tenant

bars0udin:
### The failure mode

`Plug.Helpers.retrieve_from_session/3` iterates every authenticated resource for the otp_app and deletes the session key of each one it can't resolve a user for. For a multitenant resource with `global? false`, a request that carries no tenant fails with `Ash.Error.Invalid.TenantRequired` — the query is rejected by `Ash.Actions.Read.validate_multitenancy/1` before it reaches the data layer — and the perfectly good session is deleted anyway.

### Who it affects

Any app with a tenant-scoped authenticated resource plus a pipeline that doesn't set an Ash tenant: an admin/staff area, a webhook endpoint, a health check, or simply a second authenticated resource whose pipeline is tenant-less. Users of the multitenant resource are silently signed out by requests to those routes.

Note that `retrieve_from_session/3` reads the tenant from `Ash.PlugHelpers.get_tenant(conn)` only, so this bites whenever the plug pipeline hasn't set one — `assign_new_resources/4` has a `session["tenant"]` fallback, but the plug path does not.

### Reproduction

Two authenticated resources, one of them multitenant and not `global?`; a request through a pipeline that sets no tenant. The multitenant resource's session key is gone afterwards even though the subject is untouched. The added test uses the existing `Example.MultiTenantUserWithWebAuthn` fixture, which is already `multitenancy do strategy :context end`.

### The fix

Only delete the session key when the lookup was actually able to run. If the resource declares multitenancy, isn't `global?`, and this request has no tenant, the failure says nothing about the session, so leave it alone. Genuinely absent or unresolvable subjects are still cleared, as covered by the second added test.

### DIFF
--- a/lib/ash_authentication/plug/helpers.ex
+++ b/lib/ash_authentication/plug/helpers.ex
@@ -216,12 +216,13 @@ defmodule AshAuthentication.Plug.Helpers do
         authenticate_resource_from_session(resource, session, otp_app, opts),
         resource,
         options,
-        session
+        session,
+        opts
       )
     end)
   end
 
-  defp handle_session_auth_result(conn, {:ok, user}, _resource, options, session) do
+  defp handle_session_auth_result(conn, {:ok, user}, _resource, options, session, _opts) do
     current_subject_name = current_subject_name(options.subject_name)
 
     # Restore authentication metadata from the session onto the user
@@ -230,7 +231,7 @@ defmodule AshAuthentication.Plug.Helpers do
     Conn.assign(conn, current_subject_name, user)
   end
 
-  defp handle_session_auth_result(conn, :error, resource, options, _session) do
+  defp handle_session_auth_result(conn, :error, resource, options, _session, opts) do
     current_subject_name = current_subject_name(options.subject_name)
 
     require_token? =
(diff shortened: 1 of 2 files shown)


jimsynz:
The diagnosis is correct — with a non-global multitenancy strategy and no tenant on the conn, `subject_to_user/3` is rejected by Ash's multitenancy validation before the query reaches the data layer, and the previous code treated that identically to an invalid session. I've verified your test fails against the previous implementation, and that deletion still happens when a tenant is present.

Two things before this merges:

1. Please add the inverse test: same fixture, tenant set on the conn, unresolvable subject, asserting the session key *is* deleted. That's the path that protects against the dangerous regression here — a session that should die surviving — and nothing currently pins it.

2. `tenant_required?/2` is a precondition check rather than an error match — `authenticate_resource_from_session/4` collapses every failure to `:error`, so for a tenant-scoped resource on a tenant-less request this also keeps sessions we know are dead: malformed session values, and on the `require_token_presence_for_authentication?` path an expired, tampered, or revoked token. None of these can authenticate anyone and they're cleared on the next tenanted request, so I'm not treating it as a security issue, but the comment saying "we know nothing about whether the session is valid" isn't true for those cases. Either reword it to state the trade-off, or propagate the failure reason out of `authenticate_resource_from_session/4` and keep the session only on `TenantRequired`. I'm fine with either; the latter is more precise but touches a helper shared with the LiveView path.

One caveat I'll note rather than block on: the guard reads resource-level multitenancy config, but Ash also honours action-level `multitenancy :allow_global` on the read action. A `get_by_subject` action configured that way runs fine without a tenant, and a genuine not-found there will now keep the session rather than delete it. Same hygiene-only consequence, but it's another argument for the error-match shape if you're inclined to change it.

jimsynz (approved):
tenants rights

PR #1210, opened 2026-08-31
docs: describe request action timing and account enumeration

jimsynz:
## What this documents

The magic link, password reset and OTP request actions all return a uniform `:ok`
whether or not the identity matches a user. The response *body* therefore does not
reveal which accounts exist. The response *time* is not uniform.

On a match the action mints a token and calls the configured sender inline. On a
miss it does neither. A synchronous SMTP or HTTP-API send takes tens to hundreds of
milliseconds, so the two paths are easy to separate. Account enumeration by timing
is the result.

The library's own contribution to that difference is small. Measured hit-versus-miss
for magic link is about 27 microseconds — one JWT signing operation. There is no
database write on either path by default, because `store_all_tokens?` defaults to
`false`. At that size the difference is not measurable across a network.

## Recommended mitigation

An asynchronous sender. A sender that enqueues a job and returns `:ok` immediately
keeps delivery latency off both paths and leaves only the intrinsic difference.
This is not new advice — `AshAuthentication.Sender`'s moduledoc already recommends
the same pattern for retries — so the pages cross-reference it rather than repeating
it.

## Per-strategy notes

- **Magic link** — the difference only arises when `registration_enabled?` is
  `false`. With registration enabled both paths mint a token and call the sender,
  and the measured difference falls to around 3 microseconds.
- **Password reset** — no registration gate, so the miss path always short-circuits.
- **OTP** — worst case. `Otp.generate_otp_token_for/5` and
  `generate_otp_token_for_identity/5` store the token unconditionally, regardless of
  `store_all_tokens?`, so the match path also performs a database insert the miss
  path does not. Left unquantified, as it was not measured directly.

The OTP page's request-flow step 6 previously read "never reveals user existence".
That is reworded to say the response body does not reveal it, so the page does not
contradict the new section.

## Why documentation rather than a code fix

A code fix was considered and rejected. Moving sends off the request path would
reverse the deliberate breaking change that made sender failures propagate to the
caller. It would also still leave the intrinsic difference, which the async-sender
mitigation already reduces to the same floor. The cost exceeds the size of the leak.

## Follow-up

A backport to `stable-4.0` follows separately. It covers `magic-links.md` and
`password.md` only; `otp.md` does not exist on that branch.

Documentation only. No `.ex` file is touched. `mix check --no-retry` is clean.

### DIFF
--- a/documentation/tutorials/magic-links.md
+++ b/documentation/tutorials/magic-links.md
@@ -112,3 +112,23 @@ end
 
 # ...
 ```
+
+## Request timing and account enumeration
+
+The request action returns `:ok` whether or not the identity matches a user. The
+response body therefore does not reveal which accounts exist. The response *time*
+is a different matter. When the identity matches, the action mints a token and
+calls your sender inline. When it does not match, it does neither. A synchronous
+SMTP or HTTP-API send takes tens to hundreds of milliseconds. That gap is easy to
+measure, so it separates known identities from unknown ones.
+
+The sender is the part you control, and it is the part that leaks. Make it
+asynchronous. A sender that enqueues a job and returns `:ok` immediately keeps
+delivery latency off both paths. What remains is one JWT signing operation, about
+27 microseconds. Neither path writes to the database unless you enable
+`store_all_tokens?`. A difference that small is not measurable across a network.
+See `AshAuthentication.Sender`, which recommends the same pattern for retries.
+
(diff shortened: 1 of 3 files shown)


sevenseacat (commented):
I think we need at least one example of how to make the senders asynchronous, given this update is for tutorials. Even if we don't want to assume the user is using Oban, so we have something as basic as wrapping the `Mailer.deliver` call with `Task.start`.

PR #1227, opened 2026-09-16
fix(dynamic_oidc): preserve runtime connection and email trust policy

Alt-iOS:
When migrating to the new DynamicOIDC, I found some issues, this PR Fixes dynamic OIDC’s inherited email-trust settings and preserve the selected connection when saving identities. It also adds five regression tests; all 789 checks pass.


jimsynz:
Thank you for this, and for finding it independently — you reached the same diagnosis we had under embargo, from a real migration rather than from a report.

Both problems you identified are now fixed and released, in `5.0.0-rc.14`:

- The connection is carried through to the identity write, so each `dynamic_oidc` connection gets its own identity namespace. This is published as [GHSA-73j9-m294-fvv9](https://github.com/team-alembic/ash_authentication/security/advisories/GHSA-73j9-m294-fvv9), and your reading of it was correct — two IdPs issuing the same `sub` resolved to one local user.
- `on_untrusted_email_match` and `trust_email_verified?` were advertised by the `dynamic_oidc` DSL while the struct had neither field, so Spark dropped them on `struct/2` and setting either did nothing.

The shipped change covers more ground than this PR, which is the only reason I am closing it rather than merging: it also restores the connection id at the three other sites that rebuild the strategy from the compile-time DSL — `OAuth2.IdentityChange`, `OAuth2.SignInPreparation` and the confirmation add-on — and refuses the write outright when no connection id is present, rather than falling back to the bare strategy name. Coordinating that as one embargoed change across both release lines is why this sat open while the advisory was in preparation, and I am sorry for the silence; it was not a comment on the work.

If you run `dynamic_oidc` in production, note the upgrade step in `documentation/topics/upgrading.md`: identity rows written before this change carry the bare strategy name and no longer match, so they need moving into their connection namespace with an `UPDATE` rather than being deleted.

Closing as fixed. Genuinely appreciated.

PR #1228, opened 2026-09-20
Logout everywhere notifications

Alt-iOS:
The logout everywhere action doesn't enable notifications, this changes enables it and adds a test for it. Logout everywhere without notifications enabled, would allow an already mounted liveview to still persist the user.


jimsynz (approved):
notify

